PR #7, opened 2019-03-21
Fix command name and arguments in README

rohanpm:
The command and some argument names were incorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,14 @@ ubi repos and required action is started.
 
 # Cli usage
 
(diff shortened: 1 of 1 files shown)


rohanpm:
OK, I've fixed some wrong argument names too.

About the missing arguments, I don't want to add those, I don't think it should be required for the README to list every possible argument.  What I'd rather do is, once the sphinx docs are working, try integrating https://sphinx-argparse.readthedocs.io/en/stable/ ... this generates docs using the help text defined in the argparser, so it's not necessary to duplicate the info.

PR #8, opened 2019-03-21
Get docs working via Travis & github pages

rohanpm:
Basic skeleton of docs are in place.

Currently, the classes/methods appearing in "API reference" should
be considered as examples only.  I'm not sure exactly what should
go in there as it still seems a bit unclear what is and isn't
public API.  The main goal here is to get the doc building toolchain
up and running.


rohanpm:
A GITHUB_TOKEN will need to be added to Travis CI env vars before the push from master can work.

rohanpm:
Per discussion in https://github.com/release-engineering/ubi-population-tool/issues/9, it might make sense to drop all the references to API for now.

docs/conf.py
```diff
@@ -0,0 +1,202 @@
...
+#
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'ubi-population-tool'
+copyright = '2018, Red Hat'
```
alexandrevicenzi: 2019 right?

rohanpm: Yep, fixed.

rbikar:
Token added to travis, merging.

PR #35, opened 2019-04-02
Refactor associations and unassociations, add test coverage, renamed some attributes

rbikar:
This PR is a refactor of creation of units that should associated and unassociated where I added more abstraction which should result better readability and maintainability. 
Renamed **out_repo_set** attr to **repos**, of UbiPopulateRunner class. I found it too long.

Also added tests to improve code coverage.

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -133,8 +133,8 @@ def test_match_packages(mock_ubipop_runner):
     mock_ubipop_runner.pulp.search_rpms.return_value = [get_test_pkg(name=package_name)]
     mock_ubipop_runner._match_packages()
 
-    assert len(mock_ubipop_runner.out_repo_set.packages) == 1
-    assert mock_ubipop_runner.out_repo_set.packages[package_name][0].name == package_name
+    assert len(mock_ubipop_runner.repos.packages) == 1
+    assert mock_ubipop_runner.repos.packages[package_name][0].name == package_name
 
 
 def test_match_modules(mock_ubipop_runner):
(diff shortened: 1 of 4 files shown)


tests/test_ubipop.py
```diff
@@ -308,3 +316,137 @@ def test_create_output_file(mock_ubipop_runner, mock_get_repo_pairs, mocked_ubic
...
+    modules, rpms, srpms, debug_rpms = unassociations
+    assert len(modules.units) == 0
+    assert len(rpms.units) == 0
+    assert len(srpms.units) == 0
+    assert len(debug_rpms.units) == 0
+
+
+@pytest.fixture()
+def set_logging():
+    _LOG = logging.getLogger("ubipop")
```
alexandrevicenzi: not global, I guess no need to uppercase and _ it.

ubipop/__init__.py
```diff
@@ -221,55 +223,70 @@ def _parse_blacklist_config(self):
...
+    def _determine_pulp_actions(self, units, current, diff_f):
+        if isinstance(units, dict):
+            expected = list(chain.from_iterable(units.values()))
+        else:
+            expected = units
+        to_associate = diff_f(expected, current)
+        to_unassociate = diff_f(current, expected)
+        return to_associate, to_unassociate
+
+    def _gets_action_for_mds(self, modules, current):
```
rohanpm: Naming here is a bit weird:

- this one is called *gets* action *for* mds
- next one is called *get* action pkgs (no "s", no "for")

Those kind of unnecessary inconsistencies make the code harder to work with.

rbikar: fixed

tests/test_ubipop.py
rbikar: fixed

PR #36, opened 2019-04-03
Fix handling packages of debug repos

rbikar:
Now pkgs from debug repos are handled the same way as binary rpms - whitelistiing and blacklsiting.


tests/test_ubipop.py
```diff
@@ -124,27 +125,61 @@ def test_sort_modules(mock_ubipop_runner):
...
+def test_match_binary_rpms(mock_ubipop_runner):
+    package_name = 'foo-pkg'
+    mock_ubipop_runner.pulp.search_rpms.side_effect = _get_search_rpms_side_effect(package_name)
+    mock_ubipop_runner._match_binary_rpms()
 
     assert len(mock_ubipop_runner.repos.packages) == 1
     assert mock_ubipop_runner.repos.packages[package_name][0].name == package_name
 
 
+def test_match_debug_repms(mock_ubipop_runner):
```
rohanpm: repms => rpms

rbikar: fixed

PR #38, opened 2019-04-03
Add option to write all output repos to file

rbikar:
New cli option intriduced --output-all-repos.
If provided along with --output-changed-repos, it will
write to output file all output repos.

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -17,9 +17,22 @@ TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), './data')
 
 @pytest.fixture()
 def ubi_repo_set():
-    rhel_repo_set = RepoSet("foo-rpms", "foo-source", "foo-debug")
-    ubi_repo_set = RepoSet("ubi-foo-rpms", "ubi-foo-source", "ubi-foo-debug")
-    yield UbiRepoSet(rhel_repo_set, ubi_repo_set)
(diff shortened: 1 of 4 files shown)


tests/test_ubipop.py
```diff
@@ -56,6 +69,18 @@ def get_test_mod(**kwargs):
                   kwargs.get('profiles', ''))
 
 
+def test_get_output_repo_ids(ubi_repo_set):
+    repo_ids = ubi_repo_set.get_output_repo_ids()
+    assert len(repo_ids) == 3
```
midnightercz: not sure if this makes sense when you test value directly one line bellow

tests/test_ubipop.py
```diff
@@ -56,6 +69,18 @@ def get_test_mod(**kwargs):
...
+    assert repo_ids == set(["ubi-foo-rpms", "ubi-foo-source", "ubi-foo-debug"])
+
+
+def test_get_output_repo_ids_no_debug(ubi_repo_set_no_debug):
+    repo_ids = ubi_repo_set_no_debug.get_output_repo_ids()
+    assert len(repo_ids) == 2
```
midnightercz: same here

rbikar:
@rohanpm 
Originally I thought we would like only changed repos in output file for pulp manifests. I agree, if there is no use case for that, it's not a good idea to add every possible option there. I'll leave only option for writing out all output repos.

rbikar:
I decided to consolidate these output options into one which only makes sense.

PR #42, opened 2019-04-05
Pulp action param order

rbikar:
Fixed  wrongly ordered variables the were returned from get_action() method


emilyzheng:
Association is done wrongly from dest repo to source repo:
`2019-04-08 02:40:03 +0000 [INFO    ] Associating httpd-tools-2.4.37-4.module+el8+2248+23d5e2f2.x86_64.rpm from ubi-7-server-rpms__7Server__x86_64 to rhel-7-server-rpms__7Server__x86_64
`

PR #44, opened 2019-04-05
Add integration tests.

alexandrevicenzi:
Add integration tests that will be responsible for evaluating if ubi-pop can perform operations as expected.

Tests include package and modules operations. Each test has a short description responsible for explaining what the test should do.

### DIFF
--- a/tests/integration/test_association.py
+++ b/tests/integration/test_association.py
@@ -1,0 +1,455 @@
+import os
+import subprocess
+
+from itertools import chain
+
+import pytest
+import requests
+import ubiconfig
+import ubipop
+
+from ubipop._pulp_client import Pulp
+from ubipop._utils import splitFilename as split_filename
+
+
+PULP_HOSTNAME = os.getenv('TEST_PULP_HOSTNAME')
+PULP_USER = os.getenv('TEST_PULP_USER')
+PULP_PWD = os.getenv('TEST_PULP_PWD')
+PULP_CERT_PATH = os.getenv('TEST_PULP_CERT_PATH')
+PULP_SECURE = os.getenv('TEST_PULP_SECURE', '0').lower() in ['true', 'yes', '1']
+PULP_RPM_REPO_PREFIX = '/pulp/rpmrepos/'
+
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), './data')
+
+INTEGRATION_NOT_SETUP = PULP_HOSTNAME is None
+
+
+def load_ubiconfig(filename):
+    loader = ubiconfig.get_loader(TEST_DATA_DIR)
+    return loader.load(filename)
+
+
+def run_ubipop_tool(config_file, workers=10, dry_run=False):
+    if PULP_CERT_PATH is None:
+        auth = (PULP_USER, PULP_PWD)
+    else:
+        auth = (PULP_CERT_PATH,)
+
+    up = ubipop.UbiPopulate(
+        pulp_hostname=PULP_HOSTNAME,
+        pulp_auth=auth,
+        dry_run=dry_run,
+        ubiconfig_filename_list=[config_file],
+        ubiconfig_dir_or_url=TEST_DATA_DIR,
+        insecure=not PULP_SECURE,
+        workers_count=workers,
+    )
+
+    up.populate_ubi_repos()
+
+
+def get_repos_from_cs(cs, skip_dot_version=False):
+    p = Pulp(PULP_HOSTNAME, (PULP_USER, PULP_PWD), not PULP_SECURE)
+
+    ret = p.do_request('post', 'repositories/search/', {
+        'criteria': {
+            'filters': {'notes.content_set': cs},
+        },
+        'distributors': False,
+    })
+
+    ret.raise_for_status()
+
+    for item in ret.json():
+        notes = item['notes']
+
+        if skip_dot_version and '.' in notes['platform_full_version']:
+            continue
+
+        yield {
+            'id': item['id'],
+            'name': item['display_name'],
+            'url': notes['relative_url'],
+            'arch': notes['arch']
+        }
+
+
+def query_pulp_rpms(repo_id, name=None, arch=None):
+    p = Pulp(PULP_HOSTNAME, (PULP_USER, PULP_PWD), not PULP_SECURE)
+
+    url = 'repositories/{REPO_ID}/search/units/'.format(REPO_ID=repo_id)
+
+    criteria = {'type_ids': ['rpm', 'srpm']}
+    filters = {'filters': {'unit': {}}}
+
+    if name:
+        filters['filters']['unit']['name'] = name
+        criteria.update(filters)
+
+    if arch:
+        filters['filters']['unit']['arch'] = arch
+        criteria.update(filters)
+
+    res = p.do_request('post', url, {'criteria': criteria})
+    res.raise_for_status()
+    rpms = []
+
+    for item in res.json():
+        metadata = item['metadata']
+        rpms.append({
+            'name': metadata['name'],
+            'arch': metadata['arch'],
+            'filename': metadata['filename'],
+        })
+
+    return rpms
+
+
+def query_pulp_modules(repo_id, name=None, stream=None):
+    p = Pulp(PULP_HOSTNAME, (PULP_USER, PULP_PWD), not PULP_SECURE)
+
+    url = 'repositories/{REPO_ID}/search/units/'.format(REPO_ID=repo_id)
+
+    criteria = {'type_ids': ['modulemd']}
+
+    if name and stream:
+        criteria.update({'filters': {'unit': {'name': name, 'stream': stream}}})
+
+    res = p.do_request('post', url, {'criteria': criteria})
+    res.raise_for_status()
+    modules = []
+
+    for item in res.json():
+        metadata = item['metadata']
+        modules.append({
+            'name': metadata['name'],
+            'arch': metadata['arch'],
+            'stream': metadata['stream'],
+            'artifacts': metadata['artifacts'],
+            'profiles': metadata['profiles'],
+        })
+
+    return modules
+
+
+def query_repo_rpms(query, repo_id, repo_url, force_refresh=True, arch_list=None):
+    args = [
+        'yum',
+        'repoquery',
+        '-q',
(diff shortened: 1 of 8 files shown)


alexandrevicenzi:
@rbikar @emilyzheng could you give feedback on this?

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+
+def query_pulp_rpms(repo_id, name=None, arch=None):
+    p = Pulp(PULP_HOSTNAME, (PULP_USER, PULP_PWD), not PULP_SECURE)
+
+    url = 'repositories/{REPO_ID}/search/units/'.format(REPO_ID=repo_id)
+
+    criteria = {'type_ids': ['rpm', 'srpm']}
+    filters = {'filters': {'unit': {}}}
```
danrodrig:
```
unit_filter = {}
criteria = {
    'type_ids': ['rpm', 'srpm'],
    'filters': {'unit': unit_filter},
}
```
Then you only need:
```
if name:
    unit_filter['name'] = name

if arch:
    unit_filter['arch'] = arch
```
And would be more clear to define criteria after setting name and arch in the unit_filter.

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+@pytest.mark.skipif(INTEGRATION_NOT_SETUP, reason='Integration test is not set up.')
+def test_ubipop_can_clean_repos():
+    '''
+    Test if an empty (no packages and modules) config file
+    cleans up all packages and modules in the repo.
+
+    Running this should not associate any packages or modules,
+    and also unassociate all packages and modules if any available.
```
danrodrig: neither unassociate...

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,315 @@
...
+
+        rpms = query_repo_rpms(None, 'ubi', repo_url)
+        assert len(rpms) == 0, 'Repository "{}" is not empty.'.format(repo_url)
+
+        modules = query_pulp_modules(repo_id)
+        assert len(modules) == 0, 'Repository is not empty: {}'.format(repo_id)
+
+
+@pytest.mark.skipif(INTEGRATION_NOT_SETUP, reason='Integration test is not set up.')
+def test_ubipop_can_whitelist_pkg_in_repos():
```
danrodrig: test_ubipop_can_associate_pkg_in_repos?

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,315 @@
...
+    '''
+    Test if ubipop can associate packages in the repo.
+
+    One or more packages will be published to RPM, SRPM and debug repos.
+    No packages are blacklisted.
+    No modules are whitelisted or blacklisted.
```
danrodrig: There is no blacklist for modules

alexandrevicenzi: There's no explicit blacklist but I get your point.

tests/test_ubipop_pulp_integration.py
alexandrevicenzi: This piece was based on the lib code, but I can change it.

alexandrevicenzi:
@JayZ12138 didn't get your point on QE, I'm part of AQE.

I prefer to leave the tests where they are, makes no sense to put under tests/ci, every test will be run by some CI, Travis or Jenkins.

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,440 @@
...
+        ubiconfig_filename_list=[config_file],
+        ubiconfig_dir_or_url=TEST_DATA_DIR,
+        insecure=not PULP_SECURE,
+        workers_count=workers,
+    )
+
+    up.populate_ubi_repos()
+
+
+def get_repos_from_cs(cs, skip_dot_version=False):
```
midnightercz: there's search_repo_by_cs in _pulp_client, why not use that?
Similarly for methods bellow

alexandrevicenzi: They do not return the data I want.

tox.ini
```diff
@@ -1,12 +1,19 @@
...
 deps=-rtest-requirements.txt
 commands=
         pytest -v {posargs}
 whitelist_externals=sh
 
+[testenv:integration]
+passenv = TEST_PULP_*
+deps=-rtest-requirements.txt
+commands=
+        pytest -v {posargs} tests/test_ubipop_pulp_integration.py
```
rohanpm: It would be a little nicer if you arranged it by directory and had this run pytest on `tests/integration`.

- it's a convention which can be used on any other projects as well
- it allows having integration tests split across multiple files

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,440 @@
...
+
+    if arch_list is not None:
+        args.append('--archlist={}'.format(','.join(arch_list)),)
+
+    if query is not None:
+        args.append(query)
+
+    out = subprocess.check_output(args, shell=False)
+
+    return [pkg.split(':')[1] + '.rpm' for pkg in out.split('\n') if pkg]
```
emilyzheng: When repo is empty(), it gets 'TypeError: a bytes-like object is required, not 'str''.

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,440 @@
...
+    if force_refresh:
+        args.append('--refresh')
+
+    if arch_list is not None:
+        args.append('--archlist={}'.format(','.join(arch_list)),)
+
+    if query is not None:
+        args.append(query)
+
+    out = subprocess.check_output(args, shell=False)
```
emilyzheng: yum repoquery doesn't list packages from modules:
```
$ yum repoquery -q --envra --repoid=ubi --repofrompath=ubi,http://10.0.148.48/pulp/rpmrepos/content/public/ubi/dist/ubi/server/7/7Server/x86_64/os --refresh
0:389-ds-base-1.3.1.6-10.el7.x86_64

$ repoquery -a --repofrompath=ubi,http://10.0.148.48/pulp/rpmrepos/content/public/ubi/dist/ubi/server/7/7Server/x86_64/os --repoid=ubi
389-ds-base-0:1.3.1.6-10.el7.x86_64
perl-FCGI-1:0.78-10.module+el8+2536+5485d7b4.x86_64
```

tox.ini
```diff
@@ -1,12 +1,19 @@
...
+whitelist_externals=yum
```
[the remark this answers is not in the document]
alexandrevicenzi: No, only on Jenkins for now.

tests/test_ubipop_pulp_integration.py
```diff
@@ -0,0 +1,440 @@
...
+    res.raise_for_status()
+    rpms = []
+
+    for item in res.json():
+        metadata = item['metadata']
+        rpms.append({
+            'name': metadata['name'],
+            'arch': metadata['arch'],
+            'filename': metadata['filename'],
+            'sourcerpm_filename': metadata['sourcerpm'],
```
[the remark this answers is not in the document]
alexandrevicenzi: I guess this field is not even used, will check.

tests/test_ubipop_pulp_integration.py
alexandrevicenzi: What python version are you using? I didn't get any error like this.

tox.ini
alexandrevicenzi: Yes, forgot about that, will change it.

tests/test_ubipop_pulp_integration.py
alexandrevicenzi: Just noticed that the function changed in py3, and I was testing with py2.7.

tests/test_ubipop_pulp_integration.py
alexandrevicenzi: At least on Fedora, both commands return the same data.

rohanpm: I think the problem is that if you are using repoquery from dnf, it's "module-aware" and only shows modular RPMs if they belong to a module enabled on the host system, while older repoquery from prior to modules simply lists all RPMs.

It may be that repoquery is no longer usable as a reliable way of listing all RPMs from a given repo.

tests/integration/test_association.py
```diff
@@ -0,0 +1,446 @@
...
+
+    repos = {
+        'rpm': get_repos_from_cs(cfg.content_sets.rpm.output, skip_dot_version=True),
+        'srpm': get_repos_from_cs(cfg.content_sets.srpm.output, skip_dot_version=True),
+        'debug': get_repos_from_cs(cfg.content_sets.debuginfo.output, skip_dot_version=True),
+    }
+
+    for repo_type, repo_iter in repos.items():
+        if repo_type == 'debug':
+            filter_debug = lambda pkg: 'debuginfo' in pkg.name
```
emilyzheng: *-debugsource rpms should also go to debug repo. If testing data is updated to include them, this case will fail by searching debugsource rpm in a wrong repo.

tests/integration/data/blacklist-pkg.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+arches:
+ - src
+ - x86_64
+
+packages:
+  include:
+  - MySQL-python.*
+  blacklist:
+  - MySQL-python-debuginfo.*
```
emilyzheng: The reason that MySQL-python-debuginfo is not associated to debug repo could be:
1. It's missing from 'include'
2. It exists in 'exclude'

In order to verify if 2 really works, MySQL-python-debuginfo should also be added to whitelist to avoid impact of 1.

alexandrevicenzi: You mean, have MySQL-python-debuginfo.* in both include and blacklist?

alexandrevicenzi: I guess would be better to have multiple YAML files and run them in the same test case, so we ensure that all possibilities work.

tests/integration/test_association.py
alexandrevicenzi: Wil check that

PR #48, opened 2019-04-08
Rework srpms output set creation

rbikar:
Reworked srpms output set creation. Now it doesn't guess filename of srpm if its filename is not referenced on binary rpm unit.
Updated tests.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -270,14 +270,14 @@ class UbiPopulateRunner(object):
             self.keep_n_latest_modules(units)
 
     def _create_srpms_output_set(self):
-        rpms = list(chain.from_iterable(self.repos.packages.values()))
+        rpms = chain.from_iterable(self.repos.packages.values())
         for package in rpms:
             if package.sourcerpm_filename is None:
                 _LOG.warning("Package %s doesn't reference its source rpm", package.name)
                 continue
 
-            self.repos.source_rpms[package.name].extend([Package(package.name,
-                                                                 package.sourcerpm_filename)])
+            self.repos.source_rpms[package.name].append(Package(package.name,
+                                                                package.sourcerpm_filename))
 
         blacklisted_srpms = self.get_blacklisted_packages(
             list(chain.from_iterable(self.repos.source_rpms.values())))
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -29,8 +29,7 @@ class UbiRepoSet(object):
         self.debug_rpms = defaultdict(list)
         self.modules = defaultdict(list)
         self.pkgs_from_modules = defaultdict(list)
-
-        self.source_rpms = []
+        self.source_rpms = defaultdict(list)
 
         self._ensure_repos_existence()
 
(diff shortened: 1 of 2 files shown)


ubipop/__init__.py
```diff
@@ -271,24 +270,23 @@ def _finalize_output_units(self, units, type_id):
...
             if package.sourcerpm_filename is None:
-                name, ver, rel, _, _ = splitFilename(package.filename)
-                package.sourcerpm_filename = "{n}-{v}-{r}.src.rpm".format(n=name, v=ver, r=rel)
+                _LOG.warning("Package %s doesn't reference its source rpm", package.name)
+                continue
 
-            self.repos.source_rpms.append(Package(package.name,
-                                                  package.sourcerpm_filename))
+            self.repos.source_rpms[package.name].extend([Package(package.name,
+                                                                 package.sourcerpm_filename)])
```
JayZ12138: .extend([Package(package.name,
                                                  package.sourcerpm_filename))	                                                                 package.sourcerpm_filename)])

I don't understand this change, seems completely unnecessary, can you explain?

ubipop/__init__.py
```diff
@@ -271,24 +270,23 @@ def _finalize_output_units(self, units, type_id):
             self.keep_n_latest_modules(units)
 
     def _create_srpms_output_set(self):
-        packages = chain.from_iterable(self.repos.packages.values())
-        for package in packages:
+        rpms = list(chain.from_iterable(self.repos.packages.values()))
```
JayZ12138: why need to make it a list here?

ubipop/__init__.py
JayZ12138: I mean, append the package instead of making it a list and then extend should work properly.

rbikar: Yeah, appending just Package is sufficient here.

ubipop/__init__.py
rbikar: Yes, you're right. It's not needed here. It iterates only once over rpms, so it will ok, to leave a chain obj.

alexandrevicenzi: My question is, why do you create a generator from a list to cast to a list again? Could just be:

`rpms = self.repos.packages.values()`

Right?

ubipop/__init__.py
alexandrevicenzi: `extend` is more time consuming than `append`.

ubipop/__init__.py
```diff
@@ -271,24 +270,23 @@ def _finalize_output_units(self, units, type_id):
...
-            self.repos.source_rpms.append(Package(package.name,
-                                                  package.sourcerpm_filename))
+            self.repos.source_rpms[package.name].extend([Package(package.name,
+                                                                 package.sourcerpm_filename)])
 
-        blacklisted = self.get_blacklisted_packages(self.repos.source_rpms)
-        self.repos.source_rpms = \
-            self._diff_packages_by_filename(self.repos.source_rpms, blacklisted)
+        blacklisted_srpms = self.get_blacklisted_packages(
+            list(chain.from_iterable(self.repos.source_rpms.values())))
```
alexandrevicenzi: Same as before, is this really needed?

ubipop/__init__.py
```diff
@@ -271,24 +270,23 @@ def _finalize_output_units(self, units, type_id):
...
+
+        for pkg in blacklisted_srpms:
+            self.repos.source_rpms.pop(pkg.name, None)
 
     def _determine_pulp_actions(self, units, current, diff_f):
-        if isinstance(units, dict):
-            expected = list(chain.from_iterable(units.values()))
-        else:
-            expected = units
+        expected = list(chain.from_iterable(units.values()))
```
alexandrevicenzi: Same as before, is this really needed?

ubipop/__init__.py
rbikar: self.repos.packages is {"key": ["v1", "v2]} so self.repos.packages.values() is list of lists and I need to flatten it.

ubipop/__init__.py
rbikar: Yes, it iterates several times over expected further in code, so generator is not enough.

ubipop/__init__.py
rbikar: Yes, in self.get_blacklisted_packages() it iterates several times over blacklisted_srpms, so generator is not enough.

PR #53, opened 2019-04-16
Added requests retry mechanism for all requests call in _pulp_client

### LINKED ISSUE #32: Add retry for Pulp requests
Requests to Pulp sometimes get connection error:
```
2019-03-28 04:42:45 +0000 [INFO    ] Publishing ubi-7-server-source-rpms__7_DOT_3__x86_64 in yum_distributor
Traceback (most recent call last):
  File "/home/tester/venv/bin/ubipop", line 11, in <module>
    load_entry_point('ubi-population-tool==0.1.4', 'console_scripts', 'ubipop')()
  File "build/bdist.linux-x86_64/egg/ubipop/cli.py", line 64, in entry_point
  File "build/bdist.linux-x86_64/egg/ubipop/cli.py", line 59, in main
  File "build/bdist.linux-x86_64/egg/ubipop/__init__.py", line 91, in populate_ubi_repos
  File "build/bdist.linux-x86_64/egg/ubipop/__init__.py", line 361, in run_ubi_population
  File "build/bdist.linux-x86_64/egg/ubipop/_pulp_client.py", line 110, in wait_for_tasks
  File "build/bdist.linux-x86_64/egg/ubipop/_pulp_client.py", line 123, in search_tasks
  File "build/bdist.linux-x86_64/egg/ubipop/_pulp_client.py", line 42, in do_request
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 476, in get
    return self.request('GET', url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 464, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 415, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', BadStatusLine("''",))
```
It would be better if requests can be retried than interrupting the population process immediately once there's any connection error.

midnightercz:
To make sure tool process isn't interrupted due network issues,
retry mechanism based on urllib3 Retry was implemented.
New configuration environment variables to customize this
were added:

- UBIPOP_HTTP_TOTAL_RETRIES
- UBIPOP_HTTP_RETRY_BACKOFF

fixes #32

### DIFF
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -1,6 +1,13 @@
+try:
+    from importlib import reload
+except ImportError:
+    pass
+
 import mock
+import os
 import six
 import sys
+
 
 try:
     from http.client import HTTPMessage
@@ -9,6 +16,7 @@ except ImportError:
 
 from requests.exceptions import HTTPError
 
+import ubipop._pulp_client
 from ubipop._pulp_client import Repo, Package, Pulp, HTTP_TOTAL_RETRIES
 
 import pytest
@@ -243,3 +251,11 @@ def test_retries(set_backoff_to_zero_fixture, mocked_getresponse, mock_pulp, sho
     else:
         with pytest.raises(HTTPError):
             getattr(mock_pulp, retry_call)(*retry_args)
+
+def test_retry_env_vars(mock_pulp):
+    os.environ['UBIPOP_HTTP_TOTAL_RETRIES'] = '100'
+    os.environ['UBIPOP_HTTP_RETRY_BACKOFF'] = '99'
+    reload(ubipop._pulp_client)
+    mock_pulp._make_session()
+    assert mock_pulp.adapter.max_retries.backoff_factor == 99
+    assert mock_pulp.adapter.max_retries.total == 100
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -251,11 +251,3 @@ def test_retries(set_backoff_to_zero_fixture, mocked_getresponse, mock_pulp, sho
     else:
         with pytest.raises(HTTPError):
             getattr(mock_pulp, retry_call)(*retry_args)
-
-def test_retry_env_vars(mock_pulp):
-    os.environ['UBIPOP_HTTP_TOTAL_RETRIES'] = '100'
-    os.environ['UBIPOP_HTTP_RETRY_BACKOFF'] = '99'
-    reload(ubipop._pulp_client)
-    mock_pulp._make_session()
-    assert mock_pulp.adapter.max_retries.backoff_factor == 99
-    assert mock_pulp.adapter.max_retries.total == 100
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -23,6 +23,8 @@ import pytest
 
 import requests_mock
 
+ORIG_HTTP_TOTAL_RETRIES = HTTP_TOTAL_RETRIES
+
 if sys.version_info <= (2, 7,):
     import requests_mock as rm
 
@@ -238,16 +240,20 @@ def make_mock_response(status, text=None):
 def test_retries(set_backoff_to_zero_fixture, mocked_getresponse, mock_pulp, should_retry, err_status_code,
                  env_retries, retry_call, retry_args, ok_response, expected_retries):
     global HTTP_TOTAL_RETRIES
-    if env_retries:
-        HTTP_TOTAL_RETRIES = env_retries
+    try:
+        if env_retries:
+            HTTP_TOTAL_RETRIES = env_retries
 
-    retries = [make_mock_response(err_status_code, 'Fake Http error')
-               for _ in range(HTTP_TOTAL_RETRIES)[:-1]] + [make_mock_response(200, ok_response)]
-    mocked_getresponse.side_effect = retries
+        retries = [make_mock_response(err_status_code, 'Fake Http error')
+                   for _ in range(HTTP_TOTAL_RETRIES)[:-1]]
+        retries.extend([make_mock_response(200, ok_response)])
+        mocked_getresponse.side_effect = retries
 
-    if should_retry:
-        getattr(mock_pulp, retry_call)(*retry_args)
-        assert len(mocked_getresponse.mock_calls) == expected_retries
-    else:
-        with pytest.raises(HTTPError):
+        if should_retry:
             getattr(mock_pulp, retry_call)(*retry_args)
+            assert len(mocked_getresponse.mock_calls) == expected_retries
+        else:
+            with pytest.raises(HTTPError):
+                getattr(mock_pulp, retry_call)(*retry_args)
+    finally:
+        HTTP_TOTAL_RETRIES = ORIG_HTTP_TOTAL_RETRIES
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -1,7 +1,15 @@
-from ubipop._pulp_client import Repo, Package, Pulp
-import pytest
+import mock
+import os
 import sys
+import re
+import httplib
+from requests.exceptions import HTTPError
 
+from ubipop._pulp_client import Repo, Package, Pulp, PulpRetryAdapter, HTTP_TOTAL_RETRIES
+
+import pytest
+
+import requests_mock
 
 if sys.version_info <= (2, 7,):
     import requests_mock as rm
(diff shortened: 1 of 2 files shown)


ubipop/_pulp_client.py
```diff
@@ -9,11 +13,27 @@
...
 
 class UnsupportedTypeId(Exception):
     pass
 
 
+class PulpRetryAdapter(requests.adapters.HTTPAdapter):
+    def __init__(self, *args, **kwargs):
+        kwargs['max_retries'] = Retry(
+            total=kwargs.get('total_retries', HTTP_TOTAL_RETRIES),
+            status_forcelist=[500, 502, 503, 504],
```
alexandrevicenzi: some 4xx also make sense here, for example:  429 Too Many Requests

ubipop/_pulp_client.py
```diff
@@ -9,11 +13,27 @@
 
 _LOG = logging.getLogger("ubipop")
 
+HTTP_TOTAL_RETRIES = os.environ.get("UBIPOP_HTTP_TOTAL_RETRIES", 10)
+UBIPOP_HTTP_RETRY_BACKOFF = os.environ.get("UBIPOP_HTTP_RETRY_BACKOFF", 1)
```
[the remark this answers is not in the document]
midnightercz: yeah, I should unify that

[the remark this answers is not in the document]
midnightercz: I took the value from existing project that we have for helping with requests to pulp. But for sure I'm willing to change to anything we agree on.

ubipop/_pulp_client.py
midnightercz: Again, these are from existing projects. But sure we can add more status codes there

tests/test_pulp.py
```diff
@@ -157,3 +165,69 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+    m = mock.MagicMock(name='MockResponse', status=status)
+    m.isclosed.side_effect = [False, True]
+    m.msg = mock.MagicMock(spec=httplib.HTTPMessage, headers=[], status=status)
+    m.read.return_value = text
+    return m
+
+
+@pytest.mark.parametrize(
+    "should_retry,err_status_code,env_retries,retry_call,retry_args,ok_response,expected_retries",
+    [
```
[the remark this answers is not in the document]
midnightercz: yes, point of testing is to test retries of existing methods. Not to test for example do_request.

ubipop/_pulp_client.py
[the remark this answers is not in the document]
rohanpm: I don't think these environment variables actually work, because os.environ.get will always return a string if the value was provided in environment.  I think you've written this in a way such that it only works in the default case, e.g.  HTTP_TOTAL_RETRIES defaults to int 10, but if I set UBIPOP_HTTP_TOTAL_RETRIES=10 in environment then the variable will instead be "10", probably leading to crashes.

As a general rule of thumb to avoid this, if you write os.environ.get(key, value), the 'value' should always be a string since that's what it'll be in the non-default case.  That forces you to handle string values properly instead of forgetting as happened here.

ubipop/_pulp_client.py
```diff
@@ -23,24 +40,29 @@ def __init__(self, hostname, auth, insecure=False):
         self.scheme = "https://"
         self.base_url = urljoin(self.scheme + hostname, self.PULP_API)
         self.session = None
+        self.adapter = None
         self.insecure = insecure
 
+    def _make_session(self):
+        self.session = requests.Session()
+        self.adapter = PulpRetryAdapter()
```
rohanpm: What's the point of storing self.adapter?  It doesn't seem to be ever read from outside of this method, should it simply be a local?

midnightercz: I left it there to make testing setup easier (setting backoff factor) - although I agree that could be done with changing env variable. And also when you would like to use ubipop as library I think it's kind off ugly to need setup env variables to change that. On the other hand I also don't like idea of forwarding those options to __init__ of Pulp class. So if user would like to set it's possible to access adapter directly and change it there.
I assume based on your reference to #57, you're pointing to fact that adapter shouldn't be shared as well.

rohanpm: Actually I didn't reference to #57 from here, I referenced the other way around.  The point there is just that `mount` on same session from multiple threads is not thread-safe based on prior bugs in our code, and since this commit introduces adapters it's making it more likely we'll hit such bug in future.

However, I think that should be handled in #57 and shouldn't hold back this pull request, and making tests easier is a valid reason to add an attribute to an internal class like this.  The only thing I'm requesting changes to is that the env vars don't work.

tests/test_pulp.py
```diff
@@ -157,3 +177,85 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+
+    if should_retry:
+        getattr(mock_pulp, retry_call)(*retry_args)
+        assert len(mocked_getresponse.mock_calls) == expected_retries
+    else:
+        with pytest.raises(HTTPError):
+            getattr(mock_pulp, retry_call)(*retry_args)
+
+def test_retry_env_vars(mock_pulp):
+    os.environ['UBIPOP_HTTP_TOTAL_RETRIES'] = '100'
```
rohanpm: These settings of env vars will leak across into other tests too.  It seems like these values will apply also to tests after this one, so the behavior of tests will depend on the order in which they were run.  Tests ought to avoid side-effects like that.

Honestly I'd suggest not testing this.  At least my understanding of these env vars is that they're an escape hatch for us, which we could set to workaround problems temporarily without needing code changes, but not something we've documented or committed to supporting long-term.  Tests have a maintenance cost and it's probably not worth it for something this trivial.

midnightercz: Ok, I'm not against

ubipop/_pulp_client.py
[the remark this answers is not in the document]
JayZ12138: maybe 5 times' enough

tests/test_pulp.py
```diff
@@ -193,3 +213,77 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+
+        # test 400 status is not retryable
+        (False, 400, 3, "search_repo_by_cs", ("",), '{}', 3),
+    ]
+)
+def test_retries(set_backoff_to_zero_fixture, mocked_getresponse, mock_pulp, should_retry, err_status_code,
+                 env_retries, retry_call, retry_args, ok_response, expected_retries):
+    global HTTP_TOTAL_RETRIES
+    if env_retries:
+        HTTP_TOTAL_RETRIES = env_retries
```
danrodrig: Would be nice to save a restore original value.

tests/test_pulp.py
```diff
@@ -193,3 +213,77 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+
+    retries = [make_mock_response(err_status_code, 'Fake Http error')
+               for _ in range(HTTP_TOTAL_RETRIES)[:-1]] + [make_mock_response(200, ok_response)]
```
danrodrig: `range(HTTP_TOTAL_RETRIES - 1)`

And the 'ok_response' could be added in the next line, would make the code more readable.

ubipop/_pulp_client.py
```diff
@@ -9,11 +12,25 @@
 
 _LOG = logging.getLogger("ubipop")
 
+HTTP_TOTAL_RETRIES = int(os.environ.get("UBIPOP_HTTP_TOTAL_RETRIES", 10))
+HTTP_RETRY_BACKOFF = int(os.environ.get("UBIPOP_HTTP_RETRY_BACKOFF", 1))
```
danrodrig: We should be more careful using environment variables. If they are defined empty the casting will fail. We can treat this case the same as if the environment variable is not defined.
Also, if there is a non integer value in the environment variable, I would go with warm and continue rather than crashing with a value error.
And about the casting to integer, backoff_factor in Retry is a float. So UBIPOP_HTTP_RETRY_BACKOFF=0.2 will raise a value error.

ubipop/_pulp_client.py
danrodrig: Maybe makes sense to make that configurable.

ubipop/_pulp_client.py
midnightercz: > We should be more careful using environment variables. If they are defined empty the casting will fail. We can treat this case the same as if the environment variable is not defined.

I think if user specify env variable to change this behavior, he's probably aware that wrong value could cause some problems. I'm not against handling input properly, but not sure if this options that are not going to be even documented anywhere and will probably serve only for internal debugging purpose, have to be idiot-proof  

> And about the casting to integer, backoff_factor in Retry is a float. So UBIPOP_HTTP_RETRY_BACKOFF=0.2 will raise a value error.

Good catch, will fix that in next commit

tests/test_pulp.py
```diff
@@ -1,8 +1,30 @@
-from ubipop._pulp_client import Repo, Package, ModuleDefaults, Pulp
+try:
+    from importlib import reload
```
alexandrevicenzi: does this import make sense? not used, if fails nothing happens

ubipop/_pulp_client.py
```diff
@@ -1,4 +1,7 @@
 import requests
+from urllib3.util.retry import Retry
+import os
```
alexandrevicenzi: imports could be sorted better

tests/test_pulp.py
```diff
@@ -193,3 +215,81 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+        # test custom number of retries
+        (True, 500, 3, "search_repo_by_cs", ("",), '{}', 3),
+
+        # test 400 status is not retryable
+        (False, 400, 3, "search_repo_by_cs", ("",), '{}', 3),
+    ]
+)
+def test_retries(set_backoff_to_zero_fixture, mocked_getresponse, mock_pulp, should_retry, err_status_code,
+                 env_retries, retry_call, retry_args, ok_response, expected_retries):
+    global HTTP_TOTAL_RETRIES
```
alexandrevicenzi: Just an idea, I don't like global keyword, the best way would be to use `pulp_client.HTTP_TOTAL_RETRIES = value`.

tests/test_pulp.py
```diff
@@ -193,3 +215,81 @@ def test_wait_for_tasks(mock_pulp, mock_search_task):
...
+    try:
+        if env_retries:
+            HTTP_TOTAL_RETRIES = env_retries
+
+        retries = [make_mock_response(err_status_code, 'Fake Http error')
+                   for _ in range(HTTP_TOTAL_RETRIES)[:-1]]
```
alexandrevicenzi: range(HTTP_TOTAL_RETRIES - 1)

tests/test_pulp.py
midnightercz: not sure what you mean

alexandrevicenzi: But where is used?

danrodrig: Looks like reload is not used.

And if we need the import for an specific version better use sys.version_info.

tests/test_pulp.py
danrodrig: I think global is fine here, the variable is used in this scope.

ubipop/_pulp_client.py
```diff
@@ -9,11 +12,25 @@
 
 _LOG = logging.getLogger("ubipop")
 
+HTTP_TOTAL_RETRIES = int(os.environ.get("UBIPOP_HTTP_TOTAL_RETRIES", 10))
+HTTP_RETRY_BACKOFF = float(os.environ.get("UBIPOP_HTTP_RETRY_BACKOFF", 1))
```
danrodrig: Personally, I think that the defined empty should be handled. But If nobody else thinks it's an issue I'm not requesting the change.

tests/test_pulp.py
midnightercz: I see, I removed that code and forgot to remove that import

PR #54, opened 2019-04-16
Fix passing modulemd type_id to associate function

rbikar:
type_id must be passed as tuple to associate function in pulp_client module


rbikar:
> Probably a good idea also to fix whatever allows this mistake to happen. e.g. if associate_units is not expected to take a string as that parameter, then make it crash if given; or implement whatever missing autotest allowed this bug to happen.
> 
> Otherwise it seems the kind of thing which can easily happen again.

I agree. Better auto tests is right option, I think.

PR #59, opened 2019-04-22
Copy modulemd_defaults units as well while populating ubi repos

### LINKED ISSUE #56: modulemd_defaults units should also be copied
ubi-population-tool is currently missing any support for copying of modulemd_defaults units from source to destination repos.

This seems like an oversight which could cause problems or confusing behavior.  modulemd_defaults define which profiles are enabled by default when activating a module.  If modulemd_defaults are present in input repos but missing from output UBI repos, it could mean that enabling a module from a UBI container gives you a different set of packages than if you enable it on a regular host.

Example of a modulemd_defaults unit:

```
    {
        "_content_type_id": "modulemd_defaults",
        "_href": "/pulp/api/v2/content/units/modulemd_defaults/f88add84-7569-4991-a5f0-f10f9ee1baf6/",
        "_id": "f88add84-7569-4991-a5f0-f10f9ee1baf6",
        "children": {},
        "name": "ruby",
        "profiles": {
            "2.5": [
                "common"
            ]
        },
        "stream": "2.5"
    }
```

If this unit is present in a repo, it means that someone using that repo and asking to enable "ruby:2.5" module without specifying profiles, will get "common" profile by default.  If this unit is missing, I'm not sure what the tools will do to select the profiles; maybe it will pick one arbitrarily or maybe it will give an error?

It seems that we should be copying these units into destination repos as well, to ensure enabling a module in a container built from a UBI gives the same result as enabling it elsewhere.

Here's a suggestion for the behavior:

If output repo contains any modules named "foo", and input repo contains any modulemd_defaults with name="foo", then those modulemd_defaults are also copied to the output repo.

JayZ12138:
If there modulemd_defaults for specific modulemd in the same repo,
then it should be copied to the ubi repos.
Ubipop now looks for modulemd_defaults units after searching modulemd
units with the same name and stream. If there's any, add it to push
queue.
This fixes #56

### DIFF
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -1,4 +1,4 @@
-from ubipop._pulp_client import Repo, Package, Pulp
+from ubipop._pulp_client import Repo, Package, ModuleDefaults, Pulp
 import pytest
 import sys
 
@@ -38,6 +38,11 @@ def mock_package():
 
 
 @pytest.fixture()
+def mock_mdd():
+    yield ModuleDefaults("virt", "rhel", {"2.6": ["common"]})
+
+
+@pytest.fixture()
 def mock_response_for_async_req():
     yield {"spawned_tasks": [{"task_id": "foo_task_id"}]}
 
@@ -57,6 +62,11 @@ def search_modules_response():
 
 
 @pytest.fixture()
+def search_module_defaults_response():
+    yield [{"metadata": {"name": "virt", "stream": "rhel",
+                         "profiles": {"rhel": ["default", "common"]}}}]
+
+@pytest.fixture()
 def mock_search_rpms(requests_mock, mock_repo, search_rpms_response):
     url = "/pulp/api/v2/repositories/{REPO_ID}/search/units/".format(REPO_ID=mock_repo.repo_id)
     requests_mock.register_uri('POST', url, json=search_rpms_response)
@@ -66,6 +76,12 @@ def mock_search_rpms(requests_mock, mock_repo, search_rpms_response):
 def mock_search_modules(requests_mock, mock_repo, search_modules_response):
     url = "/pulp/api/v2/repositories/{REPO_ID}/search/units/".format(REPO_ID=mock_repo.repo_id)
     requests_mock.register_uri('POST', url, json=search_modules_response)
+
+
+@pytest.fixture()
+def mock_search_module_defaults(requests_mock, mock_repo, search_module_defaults_response):
+    url = "/pulp/api/v2/repositories/{REPO_ID}/search/units/".format(REPO_ID=mock_repo.repo_id)
+    requests_mock.register_uri('POST', url, json=search_module_defaults_response)
 
 
 @pytest.fixture()
@@ -125,6 +141,18 @@ def test_unassociate_packages(mock_pulp, mock_unassociate, mock_repo, mock_packa
     assert task_ids[0] == "foo_task_id"
 
 
+def test_associate_module_defaults(mock_pulp, mock_associate, mock_repo, mock_mdd):
+    task_ids = mock_pulp.associate_module_defaults(mock_repo, mock_repo, [mock_mdd])
+    assert task_ids
+    assert task_ids[0] == 'foo_task_id'
+
+
+def test_unassociate_module_defaults(mock_pulp, mock_unassociate, mock_repo, mock_mdd):
+    task_ids = mock_pulp.unassociate_module_defaults(mock_repo, [mock_mdd])
+    assert task_ids
+    assert task_ids[0] == 'foo_task_id'
+
+
 def test_search_rpms(mock_pulp, mock_search_rpms, mock_repo):
     found_rpms = mock_pulp.search_rpms(mock_repo)
     assert len(found_rpms) == 1
@@ -140,6 +168,14 @@ def test_search_modules(mock_pulp, mock_search_modules, mock_repo):
     assert found_modules[0].nsvca == "foo-module:9.6:1111:foo-context:x86_64"
     assert found_modules[0].packages == ["foo-pkg"]
     assert found_modules[0].profiles == {"foo-prof": ["pkg-name"]}
+
+
+def test_search_module_defaults(mock_pulp, mock_search_module_defaults, mock_repo):
+    found_module_defaults = mock_pulp.search_module_defaults(mock_repo, name='virt', stream='rhel')
+    assert len(found_module_defaults) == 1
+    assert found_module_defaults[0].name == 'virt'
+    assert found_module_defaults[0].stream == 'rhel'
+    assert found_module_defaults[0].name_profiles == 'virt:[rhel:common,default]'
 
 
 @pytest.fixture()
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -5,8 +5,11 @@ from more_executors import Executors
 from concurrent.futures import as_completed
 from collections import namedtuple, defaultdict
 from ubipop._pulp_client import Pulp, Package
-from ubipop._utils import (splitFilename, AssociateActionModules, AssociateActionRpms,
+from ubipop._utils import (splitFilename, AssociateActionModules,
+                           AssociateActionModuleDefaults,
+                           AssociateActionRpms,
                            UnassociateActionModules,
+                           UnassociateActionModuleDefaults,
                            UnassociateActionRpms)
 from itertools import chain
 
@@ -28,6 +31,7 @@ class UbiRepoSet(object):
         self.packages = defaultdict(list)
         self.debug_rpms = defaultdict(list)
         self.modules = defaultdict(list)
+        self.module_defaults = defaultdict(list)
         self.pkgs_from_modules = defaultdict(list)
         self.source_rpms = defaultdict(list)
 
@@ -198,6 +202,21 @@ class UbiPopulateRunner(object):
                     self.repos.pkgs_from_modules[name_stream].extend(module_packages)
                     self.repos.packages[package_name].extend(module_packages)
 
+    def _match_module_defaults(self):
+        """Try to find modulemd_defaults units in the same repo with the same
+        name/stream of a modulemd.
+        """
+        fts = {}
+        for _, modules in self.repos.modules.items():
+            for md in modules:
+                fts[self._executor.submit(self.pulp.search_module_defaults,
+                                          self.repos.in_repos.rpm, md.name,
+                                          str(md.stream))] = md.name + str(md.stream)
+        for ft in as_completed(fts):
+            module_defaults = ft.result()
+            if module_defaults:
+                self.repos.module_defaults[fts[ft]].extend(module_defaults)
+
     def _match_packages(self, repo, packages_dict):
         """
         Add matching packages from whitelist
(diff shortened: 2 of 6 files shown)


JayZ12138:
Please review if the changes done is OK or not, 
I'll add more test cases later.

ubipop/_pulp_client.py
```diff
@@ -268,3 +299,39 @@ def nsvca(self):
 
     def __str__(self):
         return self.nsvca
+
+
+class ModuleDefaults(object):
+    """
+    module_defaults unit, defines which profiles are enabled bu default when activating
```
danrodrig: typo, "enabled bu default"

ubipop/_pulp_client.py
```diff
@@ -102,6 +102,21 @@ def search_modules(self, repo, name=None, stream=None):
                                   metadata['arch'], metadata['artifacts'], metadata['profiles']))
         return modules
 
+    def search_module_defaults(self, repo, name, stream):
+        url = "repositories/{REPO_ID}/search/units/".format(REPO_ID=repo.repo_id)
+        criteria = {"type_ids": ["modulemd_defaults"],
+                    "filters": {"unit": {"name": name, "stream": stream}}
```
alexandrevicenzi: not well indented

ubipop/_pulp_client.py
```diff
@@ -136,6 +151,15 @@ def _modules_query(self, modules):
 
         return query_list
 
+    def _module_defaults_query(self, module_defaults):
+        query_list = []
+        for md_d in module_defaults:
+            query_list.append({'$and': [{'name': md_d.name},
+                                        {'stream': md_d.stream},
```
alexandrevicenzi: also not standard indent

ubipop/_pulp_client.py
```diff
@@ -268,3 +299,39 @@ def nsvca(self):
...
+
+    @property
+    def name_profiles(self):
+        """
+        flatten the profles and prepend name
+        'ruby:2.5:common:unique'
+        """
+        result = self.name
+        for key in sorted(self.profiles):
+            result += ':%s:%s' %(key, ':'.join(sorted(self.profiles[key])))
```
alexandrevicenzi: missing a space here after %

ubipop/_pulp_client.py
JayZ12138: Hm, looks good to me.
Do you mean I should do it like this?
 criteria = {
    "type_ids": ["modulemd_defaults"],
    "filters": {"unit": {"name": name, "stream": stream}
}

Or, can you give an example of standard indent here?

ubipop/_pulp_client.py
JayZ12138: kk

ubipop/__init__.py
```diff
@@ -198,6 +202,21 @@ def _match_modules(self):
...
+    def _match_module_defaults(self):
+        """Try to find modulemd_defaults units in the same repo with the same
+        name/stream of a modulemd.
+        """
+        fts = {}
+        for _, modules in self.repos.modules.items():
+            for md in modules:
+                fts[self._executor.submit(self.pulp.search_module_defaults,
+                                          self.repos.in_repos.rpm, md.name,
+                                          str(md.stream))] = md.name + str(md.stream)
```
rbikar: md.stream should be str already

ubipop/_pulp_client.py
```diff
@@ -136,6 +152,16 @@ def _modules_query(self, modules):
 
         return query_list
 
+    def _module_defaults_query(self, module_defaults):
+        query_list = []
+        for md_d in module_defaults:
+            query_list.append({'$and': [
+                                    {'name': md_d.name},
+                                    {'stream': md_d.stream},
+                                    {'profiles': md_d.profiles}
```
rbikar: Why do you query also by profiles? 
name and stream should be enough

ubipop/_pulp_client.py
```diff
@@ -268,3 +301,39 @@ def nsvca(self):
...
+    def __init__(self, name, stream, profiles):
+        self.name = name
+        self.stream = stream
+        self.profiles = profiles  # a dict such as {'4.046':['common']}
+
+    def __str__(self):
+        return self.name
+
+    @property
+    def name_profiles(self):
```
rbikar: Probably introducing this isn't necessary. I think using name:stream should be unique enough. What do you think?

ubipop/__init__.py
```diff
@@ -344,6 +376,9 @@ def _get_pulp_actions(self, current_modules_ft, current_rpms_ft, current_srpms_f
     def _diff_modules_by_nsvca(self, modules_1, modules_2):
         return self._diff_lists_by_attr(modules_1, modules_2, 'nsvca')
 
+    def _diff_md_defaults_by_profiles(self, module_defaults_1, module_defaults_2):
+        return self._diff_lists_by_attr(module_defaults_1, module_defaults_2, 'name_profiles')
```
rbikar: Not sure here,  but shouldn't we do diff by name:stream only?
I think we can see situation when this won't work.
E.g.:
In ubi repo we have already modulemd_defaults unit with the same name:stream as in input repo but different profiles. What will happen? Old and also new will be in ubi repo in the end, which is not what we want I guess.

JayZ12138: I'm thinking name:profiles should be used
Take your example, in the ubi repo there's already modulemd_defaults unit with same name:stream, but with different profiles, then by the function _determine_pulp_actions (line 307), the old one will be unassociated and the new one will be copied there. This should be the right behavior we want, as if profiles changes, then it means the what enables by default should be changed for that repo, so we should replace the old modulemd_defaults unit with the new one.

JayZ12138: @rbikar

rbikar: hmm, yep, you're right. Sorry.
It won't work with just name:stream only as it won't detect difference in profiles

ubipop/_pulp_client.py
rbikar: I'm taking this back.

tests/test_pulp.py
```diff
@@ -125,6 +141,18 @@ def test_unassociate_packages(mock_pulp, mock_unassociate, mock_repo, mock_packa
     assert task_ids[0] == "foo_task_id"
 
 
+def test_associate_module_defaults(mock_pulp, mock_associate, mock_repo, mock_mdd):
+    task_ids = mock_pulp.associate_module_defaults(mock_repo, mock_repo, [mock_mdd])
+    assert len(task_ids[0])
```
danrodrig: `assert task_ids`

tests/test_pulp.py
```diff
@@ -125,6 +141,18 @@ def test_unassociate_packages(mock_pulp, mock_unassociate, mock_repo, mock_packa
...
+    assert task_ids[0] == 'foo_task_id'
+
+
+def test_unassociate_module_defaults(mock_pulp, mock_unassociate, mock_repo, mock_mdd):
+    task_ids = mock_pulp.unassociate_module_defaults(mock_repo, [mock_mdd])
+    assert len(task_ids[0])
```
danrodrig: `assert task_ids`

ubipop/_pulp_client.py
```diff
@@ -182,7 +208,8 @@ def associate_units(self, src_repo, dest_repo, units, type_ids):
     def _get_query_list(self, type_ids, units):
```
rohanpm: Hmm... guess this isn't introduced by this commit, but semantics of this method seem a bit strange.

It accepts type_ids which presumably is a list/tuple, but instead of processing everything in the list, it uses if, elif, elif to select only a single thing from the list... so for example, if type_ids is ["modulemd", "rpm"] then it behaves the same as if type_ids was just ["modulemd"].  Why is modulemd higher priority than modulemd_defaults which is higher priority than rpm/srpm?  What's the point of it being a list anyway?

I'm not asking for this to be changed here though, since it wasn't introduced in this PR.

ubipop/_pulp_client.py
```diff
@@ -268,3 +301,39 @@ def nsvca(self):
...
+        self.profiles = profiles  # a dict such as {'4.046':['common']}
+
+    def __str__(self):
+        return self.name
+
+    @property
+    def name_profiles(self):
+        """
+        flatten the profles and prepend name
+        'ruby:2.5:common:unique'
```
rohanpm: Is this a syntax you've seen somewhere else in tools dealing with modules, or did you just make it up here?

I think it's a poor choice due to sharing the same format with module specifiers themselves, typically given as N:S:V:C:A with trailing portions able to be dropped.  It's also a bit weird that : is doubling up here as "separator between different types of data" and "separator between items next to each other in a list".

So, for example, if I just have a string 'ruby:2.5:common:unique' how do I know whether that's referring to module named 'ruby', stream '2.5', version 'common', context 'unique'?  Or module named 'ruby', stream '2.5' containing profiles named 'common' and 'unique' ?

Actually, isn't this format itself ambiguous even if considered alone?  You're returning only a single string and every separator is ':', even for multiple keys under self.profiles (i.e. multiple streams).  So, correct me if I'm wrong, but it seems both of these:

```
 "name": "ruby",
 "profiles": {
    "2.5": ["common"],
    "3.1": ["default", "minimal"],
 },

 "name": "ruby",
 "profiles": {
    "2.5": ["common", "3.1", "default", "minimal"],
 },
```

...would be flattened to:

```
ruby:2.5:common:3.1:default:minimal
```

Hence this is actually an ambiguous transformation, multiple different inputs can map to the same outputs.  I'm sure the above case is not that likely, nevertheless it seems technically unsafe for use in the diffs here.

If this is just an arbitrary format you're making up here, would it be better to choose something which is unambiguous both when considered alone and when compared with the existing N:S:V:C:A scheme?

Although, going a bit further - why do we even need to be flattening this into a string for diff purposes?  If you want to be able to test whether two objects are equal, the most straightforward way to do that is to implement `__eq__` on the class and then check instance1 == instance2.  It's unnecessarily convoluted to instead introduce some flattened string formatting just so you can check instance1.string_formatted == instance2.string_formatted.  Though I didn't check whether that's also used for some logging purposes.

JayZ12138: Yeah. It's made up myself, I was trying to make something that's unique to a modulemd_defaults unit, so we can determine if it's already in the repo/need to be replaced or not. Didn't expect we should be able to tell what the unit is from that string.
I did implement __eq__ in the ModueDefaults class, then I see how it's dealt with module and rpm, so I changed to this. 
I'll do changes, either make the string more clear or use __eq__ instead.

JayZ12138: I changed the format to 'ruby:[2.5:common,unique]' now, but I still feel a little strange to have [] inside the string. Do you think it's OK? 
Or I'll change to __eq__ tomorrow.

rohanpm: Yeah, I think this change addresses the main points I'd raised.

I think the code around here is still more complicated than it needs to be - it still should be unnecessary to make up string representations of objects just to store and compare them.  That's why we have the ability to override `__hash__`, `__eq__` and similar methods in the first place, so you can compare objects, use them in dicts and so on for your own custom classes just as easily as you can do for integers, strings etc.

I understand there's some urgency to get this fixed though and I don't want to block it just for that.

PR #63, opened 2019-04-25
pulp search_module_defaults should allow name and stream is None

JayZ12138:
While searching the existing modulemd_defaults units in a repo, in
order to get all the units, we don't pass name stream arguments to
search_module_defaults method, so, it should allow these arguments
to be None

### DIFF
--- a/ubipop/_pulp_client.py
+++ b/ubipop/_pulp_client.py
@@ -103,13 +103,13 @@ class Pulp(object):
                                   metadata['arch'], metadata['artifacts'], metadata['profiles']))
         return modules
 
-    def search_module_defaults(self, repo, name, stream):
+    def search_module_defaults(self, repo, name=None, stream=None):
         url = "repositories/{REPO_ID}/search/units/".format(REPO_ID=repo.repo_id)
-        criteria = {
-            "type_ids": ["modulemd_defaults"],
-            "filters": {"unit": {"name": name, "stream": stream}}
-        }
+        criteria = {"type_ids": ["modulemd_defaults"]}
+        if name and stream:
+            criteria.update({"filters": {"unit": {"name": name, "stream": stream}}})
         payload = {"criteria": criteria}
+
         ret = self.do_request("post", url, payload)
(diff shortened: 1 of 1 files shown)


JayZ12138:
> Yes, this was unfortunately overlooked in previous PR, but @JayZ12138 would you mind adding test/s for _get_current_content method in UbiPopulateRunner?

@rbikar, hmm, I can. However, this method only returns a list of future objects, doesn't have a lot to test...
I feel we should add integration test to test the whole run_ubi_population method.. but that looks like another story:)

rbikar:
> > Yes, this was unfortunately overlooked in previous PR, but @JayZ12138 would you mind adding test/s for _get_current_content method in UbiPopulateRunner?
> 
> @rbikar, hmm, I can. However, this method only returns a list of future objects, doesn't have a lot to test...
> I feel we should add integration test to test the whole run_ubi_population method.. but that looks like another story:)

Right, but you can still test if those futures finish correctly. I mean you can still take that list of futures and check results (with as_completed or future.result())

It can be something like test_associate_units()

JayZ12138:
> Right, but you can still test if those futures finish correctly. I mean you can still take that list of futures and check results (with as_completed or future.result())
> 
> It can be something like test_associate_units()

Hmm, now I have another thought. 
The method _get_cuurent_content basically runs these search_* methods in threads, then we can just check if the right argument is passed to these methods, instead of checking the result of them. As for the result, I guess we should either ran integration test for the whole population process or the unit tests separately for search_* methods.

rbikar:
> > Right, but you can still test if those futures finish correctly. I mean you can still take that list of futures and check results (with as_completed or future.result())
> > It can be something like test_associate_units()
> 
> Hmm, now I have another thought.
> The method _get_cuurent_content basically runs these search_* methods in threads, then we can just check if the right argument is passed to these methods, instead of checking the result of them. As for the result, I guess we should either ran integration test for the whole population process or the unit tests separately for search_* methods.

Right, it's not about result of search here.

PR #67, opened 2019-04-26
Use cmp_version package for version comparison

### LINKED ISSUE #66: rpm_vercmp python package is not reliable
Currently used implementation of vercmp in rpm_vercmp package is not reliable.
E.g.
vercmp("rubygems-2.0.14-25.el7_1.noarch.rpm", "rubygems-2.0.14.1-34.el7_6.noarch.rpm")
is 1
which is not true. (the latter pkg is newer)

I already investigated different vercmp package - https://pypi.org/project/cmp_version/
which works well.

Acceptance criteria:
Use cmp_version package for comparing package version.

rbikar:
Previously used rpm-vercmp was unreliable for
some variants of version, so this commit
moves to cmp_version python package which
is more reliable.

Fix #66 .

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -124,15 +124,15 @@ def test_packages_names_by_profiles_all_profiles(mock_ubipop_runner):
 
 def test_sort_packages(mock_ubipop_runner):
     packages = [
-        get_test_pkg(filename="tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm"),
-        get_test_pkg(filename="tomcatjss-5.3.6-1.el8+1944+b6c8e16f.noarch.rpm"),
-        get_test_pkg(filename="tomcatjss-9.3.6-1.el8+1944+b6c8e16f.noarch.rpm")]
+        get_test_pkg(filename="rubygems-2.0.14.1-34.el7_6.noarch.rpm"),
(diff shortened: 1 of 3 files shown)


tests/test_ubipop.py
```diff
@@ -124,15 +124,15 @@ def test_packages_names_by_profiles_all_profiles(mock_ubipop_runner):
 
 def test_sort_packages(mock_ubipop_runner):
     packages = [
-        get_test_pkg(filename="tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm"),
-        get_test_pkg(filename="tomcatjss-5.3.6-1.el8+1944+b6c8e16f.noarch.rpm"),
-        get_test_pkg(filename="tomcatjss-9.3.6-1.el8+1944+b6c8e16f.noarch.rpm")]
+        get_test_pkg(filename="rubygems-2.0.14.1-34.el7_6.noarch.rpm"),
+        get_test_pkg(filename="rubygems-2.0.14-25.el7_1.noarch.rpm"),
+        get_test_pkg(filename="rubygems-2.0.13.1-34.el7_6.noarch.rpm")]
```
midnightercz: can you also add here package that would differ only in release?

rbikar: done

PR #69, opened 2019-04-26
Allow multi arch pkgs in one repo

### LINKED ISSUE #68: Allow multi arch pkgs in output repo
Currently we take only the latest package and put it into output repository. Is some cases there are pkgs with more arches in repository and they need to be in output repo as well, not just the latest one. But the latest one of each arch.

Acceptance criteria:
Allow multi arch pkgs in output pkgs set

rbikar:
This change will allow to have pkgs with different arches
in one output repo.
Previously it took only the latest pkg and arch
wasn't taken into account.

Also added test and updated releated one.

Fixes  #68

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -615,6 +615,7 @@ class UbiPopulateRunner(object):
         """
         Keep n latest non-modular packages,
         modular packages are kept only if they are referenced by some of remaining modules
+        Parameter packages: sorted list of Packages objects, oldest goes first
         """
         packages_to_keep = []
         non_modular_pkgs = []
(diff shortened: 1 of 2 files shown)


ubipop/__init__.py
```diff
@@ -630,4 +631,15 @@ def keep_n_latest_packages(self, packages, n=1):
             else:
                 non_modular_pkgs.append(package)
 
-        packages[:] = non_modular_pkgs[-n:] + packages_to_keep
+        # filter non-modular pkgs per arches, there can be rpms with different arches
+        # for package in one repository
+        pkgs_per_arch = defaultdict(list)
+        for pkg in non_modular_pkgs:
+            _, _, _, _, arch = splitFilename(pkg.filename)
```
alexandrevicenzi: `_, _, _, _, arch = splitFilename(pkg.filename)` is ugly, maybe `arch = splitFilename(pkg.filename)[4]`

PR #70, opened 2019-04-26
Get rid of 'profiles' in module_defaults association

JayZ12138:
Previously, 'profiles' was the key part to get the right modulemd_defaults
units to associate and unassociate. However, as the keyword in profiles
always include speical character '.'(dot), it leads to mongodb query failure.

Since while assmbling the lists of md_defaults_assoc and md_defaults_unassoc,
the 'profiles' has already been considered, we don't have to re-query with
it, but, because association and unassociation happens at the same time in
threads, it could delete the new unit if association happens before unassociation,
so, we serialize the modulemd_defaults units unassociation and association.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -363,9 +363,6 @@ class UbiPopulateRunner(object):
         associations = (AssociateActionModules(modules_assoc,
                                                self.repos.out_repos.rpm,
                                                self.repos.in_repos.rpm),
-                        AssociateActionModuleDefaults(md_defaults_assoc,
-                                                      self.repos.out_repos.rpm,
-                                                      self.repos.in_repos.rpm),
                         AssociateActionRpms(rpms_assoc,
                                             self.repos.out_repos.rpm,
                                             self.repos.in_repos.rpm),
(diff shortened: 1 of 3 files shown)


JayZ12138:
> 1. Please add test coverage.
> 2. Isn't there a way how to pass special chars to mongo query?
> 3. Otherwise it looks ok.

1. Yes, I'm adding
2. No. There isn't any way to pass '.' to query in keyword. The mongodb team is trying to support it.

ubipop/__init__.py
```diff
@@ -447,6 +453,28 @@ def _associate_unassociate_units(self, action_list):
 
         return fts
 
+    def _associate_unassociate_md_defaults(self, action_md_ass, action_md_unass):
```
danrodrig: Can you de-duplicate this?

JayZ12138: Do you mean use the _associate_unassociate_units method?
I feel that's a good idea,,, at leat can save me from writing test cases..

JayZ12138: Hm, I feel it's more clear to put these modulemd_defaults units association and unassociation in one method for me.
But I can use _associate_unassociate_units method in _associate_unassociate_md_defaults to reduce duplicates.

danrodrig: Yep, you can use _associate_unassociate_units and move the code for waiting to another method. Same code is used at least in 3 places.

ubipop/__init__.py
```diff
@@ -447,6 +453,23 @@ def _associate_unassociate_units(self, action_list):
...
+        then start new units association
+        """
+        fts_unass = self._associate_unassociate_units(action_md_unass)
+        for ft in as_completed(fts_unass):
+            tasks = ft.result()
+            if tasks:
+                self.pulp.wait_for_tasks(tasks)
+
+        fts_ass = self._associate_unassociate_units(action_md_ass)
+        for ft in as_completed(fts_ass):
```
danrodrig: This code is repeated in at least three places. Here, line 462 and line 436...

JayZ12138: hmm, interesting,, I invited cd team to review and it disappeared after you commented. 

And I see what you mean at the first time, I'm just hesitated if I should wrap them into another new method:
1. the logic is pretty simple here
2. There are already a lot of methods here and in utils, I feel it could reduce duplicates, but increases overall complicity. 

What do you think?

rohanpm:
> 2. No. There isn't any way to pass '.' to query in keyword. The mongodb team is trying to support it.

That's amazing.  I almost couldn't believe it, but it's right there in 
 https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names, "The MongoDB Query Language cannot always meaningfully express queries over documents whose field names contain these characters".  So Pulp is saving modulemd_defaults units which can't possibly be searched over, it's crazy.

ubipop/__init__.py
rohanpm: I think it'd be worthwhile, e.g.

```
fts_unass = self._associate_unassociate_units(action_md_unass)
self._await_pulp(fts_unass)
```

Where await_pulp is defined as "await futures; if they complete and return pulp tasks, then await those tasks; or raise if anything fails".

Longer term (too big to include in this commit), I think the Pulp tasks themselves should be represented as futures.

PR #73, opened 2019-04-30
Should pass iterable to _associate_unassociate_units method

JayZ12138:
Now since the modulemd_defaults units association and unassociation
actions are being done separately and _associate_unassociate_units
method only takes iterable as input, we need to make them iterable
first.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -436,7 +436,7 @@ class UbiPopulateRunner(object):
             # wait for associate/unassociate tasks
             self._wait_pulp(fts)
 
-            self._associate_unassociate_md_defaults(mdd_association, mdd_unassociation)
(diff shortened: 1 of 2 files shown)


ubipop/__init__.py
```diff
@@ -454,10 +454,10 @@ def _associate_unassociate_md_defaults(self, action_md_ass, action_md_unass):
         Unassociate old module defaults units first, wait until done and
         then start new units association
         """
-        fts_unass = self._associate_unassociate_units(action_md_unass)
+        fts_unass = self._associate_unassociate_units((action_md_unass,))
```
danrodrig: You should pass the iterable when calling _associate_unassociate_md_defaults.

alexandrevicenzi:
This was already merged by #78.

PR #78, opened 2019-05-02
Fix type error.

alexandrevicenzi:
Now since the modulemd_defaults units association and unassociation
actions are being done separately and _associate_unassociate_units
method only takes iterable as input, we need to make them iterable
first.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -426,17 +426,24 @@ class UbiPopulateRunner(object):
                                    current_debug_rpms_ft)
 
         if self.dry_run:
-            self.log_curent_content(current_modules_ft, current_module_defaults_ft,
-                                    current_rpms_ft, current_srpms_ft, current_debug_rpms_ft)
-            self.log_pulp_actions(associations+mdd_association,
-                                  unassociations+mdd_unassociation)
+            self.log_curent_content(
+                current_modules_ft,
+                current_module_defaults_ft,
+                current_rpms_ft,
+                current_srpms_ft,
+                current_debug_rpms_ft,
+            )
+            self.log_pulp_actions(
+                associations + (mdd_association,),
+                unassociations + (mdd_unassociation,),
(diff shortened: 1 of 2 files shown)


alexandrevicenzi:
This is the same patch as #73 plus fixes.

ubipop/__init__.py
```diff
@@ -426,17 +426,24 @@ def run_ubi_population(self):
...
-                                  unassociations+mdd_unassociation)
+            self.log_curent_content(
+                current_modules_ft,
+                current_module_defaults_ft,
+                current_rpms_ft,
+                current_srpms_ft,
+                current_debug_rpms_ft,
+            )
+            self.log_pulp_actions(
+                associations + (mdd_association,),
```
midnightercz: I think it would be better to fix _get_upload_actions to return unified result, i.e. tuple of lists instead of tuple of mixed types. then here would be mdd_associations without no need of wrapping it in tuple

ubipop/__init__.py
```diff
@@ -426,17 +426,24 @@ def run_ubi_population(self):
                                    current_debug_rpms_ft)
 
         if self.dry_run:
-            self.log_curent_content(current_modules_ft, current_module_defaults_ft,
-                                    current_rpms_ft, current_srpms_ft, current_debug_rpms_ft)
-            self.log_pulp_actions(associations+mdd_association,
-                                  unassociations+mdd_unassociation)
+            self.log_curent_content(
```
midnightercz: note that, this coding style change is not related to the issue

alexandrevicenzi: Yeap, but as I change some lines nearby felt more readable to change as well.

ubipop/__init__.py
alexandrevicenzi: As this was from #73 I just left the same code, I only fixed one issue that was not working.

PR #83, opened 2019-05-03
Fix linter issues

alexandrevicenzi:
- Fixed code styling
- Fixed most pylint issues

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -1,17 +1,23 @@
-import re
 import logging
-import ubiconfig
-from more_executors import Executors
-from concurrent.futures import as_completed
+import re
+
 from collections import namedtuple, defaultdict
-from ubipop._pulp_client import Pulp, Package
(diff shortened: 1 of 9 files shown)


ubipop/__init__.py
```diff
@@ -163,7 +169,7 @@ def _get_repo_counterpart(self, input_repo, repos_to_match):
             if input_repo.arch == repo.arch and \
                     input_repo.platform_full_version == repo.platform_full_version:
                 return repo
-
+        return None
```
danrodrig: Notice that we are no longer enforcing inconsistent-return-statements

rbikar: This also a reason for failing test coverage check.

alexandrevicenzi:
@rbikar I try to squash to not pollute the history, but good point about review.

rbikar:
> @rbikar I try to squash to not pollute the history, but good point about review.

I always squash before merging to master (there is an option for it: squash and merge). So history is not polluted.

PR #84, opened 2019-05-03
Include all module packages if no profile is defined

### LINKED ISSUE #81: All packages from artifacts should be copied if profiles' not specified
**Description:**

Currently, while matching the modules, it adds the modular packages to repo.packages and then copied to destination repos. Depends on if profiles is specified in the config file or not, there are two different cases:
1. If profiles is specified in config file, then only the packages specified in the profiles will be added and copied to destination repo, this is right. 
2. If profiles is not specified in the config file, then it uses the profiles from the module metadata and copies the modular packages shown there. This isn't the behavior we want, the expected behavior is copy all the modular packages from 'artifacts' in the module metadata.

The functions related to this error are  '_match_modules', 'get_packages_names_by_profiles' and 'get_packages_from_module'

**Acceptance Criteria:**

if profiles is not defined in the config file, then all packages from 'artifact' in module metadata should be copied to ubi repo.

danrodrig:
Packages for modules are included based on the package name present in
the repositories module profiles. This leads to missing packages in the
output repositories, as not every package required is included in module
profiles.

In this change, if no module profile is defined in the UBI
configuration, include every package present in the module artifacts.

Fix #81

### DIFF
--- a/tests/data/ubiconf_golang.yaml
+++ b/tests/data/ubiconf_golang.yaml
@@ -1,0 +1,52 @@
+content_sets:
+  rpm:
+    output: ubi-8-for-x86_64-appstream-rpms
+    input: rhel-8-for-x86_64-appstream-rpms
+  srpm:
+    output: ubi-8-for-x86_64-appstream-source-rpms
+    input: rhel-8-for-x86_64-appstream-source-rpms
+  debuginfo:
+    output: ubi-8-for-x86_64-appstream-debug-rpms
+    input: rhel-8-for-x86_64-appstream-debug-rpms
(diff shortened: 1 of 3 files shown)


JayZ12138:
Is there a way to skip the coverage check? currently, our tests doesn't make much sense, in my opinion, no need to add these less meaningful test cases.

danrodrig:
> Is there a way to skip the coverage check? currently, our tests doesn't make much sense, in my opinion, no need to add these less meaningful test cases.

Better to keep the rule and tests, I guess.

PR #85, opened 2019-05-03
Include test code in linter.

alexandrevicenzi:
Tests should also be linted by pylint
because it's also part of the library.

Docs could also be added but I felt that this
does not belong to the library codebase.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -6,8 +6,8 @@ import pytest
 from ubipop.cli import main
 
 
-@pytest.fixture()
-def mock_ubipopulate():
+@pytest.fixture(name='mock_ubipopulate')
+def fixture_mock_ubipopulate():
     with mock.patch('ubipop.UbiPopulate') as mocked_ubipopulate:
         yield mocked_ubipopulate
 
--- a/tests/test_pulp.py
+++ b/tests/test_pulp.py
@@ -10,28 +10,27 @@ import six
 
 from mock import MagicMock, patch
 from requests.exceptions import HTTPError
-from ubipop._pulp_client import Repo, Package, ModuleDefaults, Pulp, HTTP_TOTAL_RETRIES
+from ubipop import _pulp_client as pulp_client
+from ubipop._pulp_client import Repo, Package, ModuleDefaults, Pulp
 
-ORIG_HTTP_TOTAL_RETRIES = HTTP_TOTAL_RETRIES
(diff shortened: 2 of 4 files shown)


alexandrevicenzi:
Travis check will probably fail with this change, but #83 should have fixed most issues.

tox.ini
```diff
@@ -22,7 +22,7 @@ deps=
 	pylint
 	-rtest-requirements.txt
 commands=
-	sh -c 'pylint ubipop; test $(( $? & (1|2|4|32) )) = 0'
+	sh -c 'pylint -f colorized ubipop tests/*.py tests/**/*.py; test $(( $? & (1|2|4|32) )) = 0'
```
rohanpm: Seems like just "tests" should work rather than having to glob for .py files?

rohanpm:
It'll need to be passing before merge of course.

tox.ini
alexandrevicenzi: tests is not a module, so to add only tests we need to include init file in it.

litleleprikon:
Probably commits should be rearranged?

tests/test_pulp.py
```diff
@@ -229,40 +237,42 @@ def test_search_modules(mock_pulp, mock_search_modules, mock_repo):
...
+@pytest.fixture(name='mocked_getresponse')
+def fixture_mocked_getresponse():
     with patch("urllib3.connectionpool.HTTPConnectionPool._get_conn") as mocked_get_conn:
         yield mocked_get_conn.return_value.getresponse
 
 
-@pytest.fixture()
-def set_backoff_to_zero_fixture(mock_pulp):
+@pytest.fixture(name='set_backoff_to_zero_fixture')
+def fixture_set_backoff_to_zero_fixture(mock_pulp):
```
midnightercz: fixture_ * _fixture ?

tests/test_cli.py
```diff
@@ -6,8 +6,8 @@
 from ubipop.cli import main
 
 
-@pytest.fixture()
-def mock_ubipopulate():
+@pytest.fixture(name='mock_ubipopulate')
+def fixture_mock_ubipopulate():
```
midnightercz: What lint warning was fixed here?

alexandrevicenzi: Name redefined

tests/test_pulp.py
alexandrevicenzi: Nice catch

Issue #52, opened 2019-04-12
Source rpm of modules package not in whitelist cannot be handled


emilyzheng:
There's httpd modulemd which has profiles:
```
    "profiles" : {
        "default" : [ 
            "httpd-tools"
        ]
    }
```
When package httpd-tools included in modulemd is in packages whitelist:
```
packages:
  include:
  - httpd-tools.*
modules:
  include:
  - name: httpd
    stream: 2.4
    profiles:
    - default
```
Population log is as below ('sourcerpm' of httpd-tools-2.4.6-88.el7.x86_64.rpm is httpd-2.4.6-88.el7.src.rpm):
```
2019-04-12 02:49:21 +0000 [WARNING ] Package httpd-tools doesn't reference its source rpm
...
2019-04-12 02:49:21 +0000 [INFO    ] Would associate httpd-tools-2.4.6-88.el7.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
2019-04-12 02:49:21 +0000 [INFO    ] Would associate httpd-tools-2.4.6-88.el7.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
2019-04-12 02:49:21 +0000 [INFO    ] Would associate httpd-2.4.6-88.el7.src.rpm from rhel-7-server-source-rpms__7Server__x86_64 to ubi-7-server-source-rpms__7Server__x86_64
```
Issues:
1. httpd-tools-2.4.6-88.el7.x86_64.rpm is associated twice because it's counted from both whitelist and modulemd.
2. Packages counted from modulemd don't have their sourcerpm searched, so there's warning 'Package httpd-tools doesn't reference its source rpm'.

midnightercz:
for 1) isn't that captured also in https://github.com/release-engineering/ubi-population-tool/issues/74 ? Can we then reduce scope of this issue just to 2) ?

rbikar:
This issue was partially fixed:
Source rpms of modular pkgs were fixed in #91
Issue with duplicated associations is probably still valid but I think that it's connected to #94

Issue #27, opened 2019-03-27
No check on existence of packages in modules


emilyzheng:
Packages which are in package whitelist and not existing in source repo won't be copied to ubi repo, but packages included in modules will be, even if they don't exist in source repo. E.g. there's ubipop log:
```
2019-03-27 05:25:32 +0000 [INFO    ] Associating httpd-tools-2.4.37-4.module+el8+2248+23d5e2f2.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
```
httpd-tools is a package in httpd module. It won't really be associated to ubi-7-server-rpms__7Server__x86_64 when it doesn't exist in rhel-7-server-rpms__7Server__x86_64.

Above log is misleading since it says the rpm will go to ubi repo, but actually it won't, although this may not be harmful to final content in ubi repo.

emilyzheng:
Similar issue on source rpm/debuginfo rpm, they may be outputted as associated to ubi repos, but not even exist in source rpm repo/debug rpm repo.

alexandrevicenzi:
@emilyzheng @rbikar it would be nice to show a warning if a package is missing, otherwise, the module is kinda useless. Let's suppose httpd module, it makes no sense to install it if mod_http2 package is missing.

Or even, the association of a module should fail if packages are missing.

rohanpm:
It sounds like the assumption there is that anything in modulemd "artifacts" must be available on one of the system's repos for the module to be usable, but that's not the case.  A simple example is that modulemd artifacts have references to debuginfo/debugsource RPMs, but users don't have debuginfo repos enabled by default (and most users will never enable them). Yet this doesn't interfere with installing or using the module.

Maybe there are also other non-debuginfo or non-source-RPM cases where it's expected for something to be listed in modulemd artifacts but not actually be present in repos enabled on the end-user system.  I don't know...

Unless you can find clear information about it somewhere I'd recommend only going as far as warnings, and not fatal errors, when a modulemd refers to an RPM which can't be found.

alexandrevicenzi:
@rohanpm I agree with you, it's difficult to ensure, so better go only with a warning.

alexandrevicenzi:
This task is related to #18.

rbikar:
Fixed by #91

PR #91, opened 2019-05-07
Reworked matching modular rpms

### LINKED ISSUE #24: Duplicated package association when profiles is None
get_packages_names_by_profiles may return a list containing duplicated package names when 'profiles' is not defined in ubi config. E.g. for modulemd unit:
```
   "profiles" : {
        "default" : [ 
            "httpd", 
            "httpd-filesystem"
        ],
        "minimal" : [ 
            "httpd"
        ],
        "devel" : [ 
            "httpd", 
            "httpd-devel"
        ]
    },
```
The same package 'httpd' would be associated for 3 times because it exists in 3 profiles. Dry-run log is:
```
2019-03-26 04:47:23 +0000 [INFO    ] Would associate httpd-2.4.37-4.module+el8+2248+23d5e2f2.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
2019-03-26 04:47:23 +0000 [INFO    ] Would associate httpd-2.4.37-4.module+el8+2248+23d5e2f2.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
2019-03-26 04:47:23 +0000 [INFO    ] Would associate httpd-2.4.37-4.module+el8+2248+23d5e2f2.x86_64.rpm from rhel-7-server-rpms__7Server__x86_64 to ubi-7-server-rpms__7Server__x86_64
```
The duplication should be avoided.

rbikar:
This commit includes more precise handling
of modular packages. Every rpm is checked againts
pulp if it exists in binary rpm or debug rpm repo
and then rpm is added accordingly.
This leads to correct assocications of debug rpms
and also correct references for source rpms
of modular packages.

Also fixed overlapping package names in profiles
which won't duplicated now.

Fixes #24 #52  #27

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -615,7 +615,7 @@ class UbiPopulateRunner(object):
                 modules to process
 
         Returns:
-            set of str:
+            list of str:
                 names of packages within matching modules & profiles
         """
         packages_names = []
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -191,22 +191,27 @@ class UbiPopulateRunner(object):
                 name_stream = fts[ft][0]
                 profiles = fts[ft][1]
                 self.repos.modules[name_stream].extend(input_modules)
-
                 if profiles:
                     # Include packages from module profiles only.
                     packages_names = self.get_packages_names_by_profiles(profiles, input_modules)
                     for package_name in packages_names:
                         module_packages = self.get_packages_from_module(input_modules,
                                                                         package_name)
+
+                        rpms, debug_rpms = module_packages
                         # for reference which pkgs are from modules
-                        self.repos.pkgs_from_modules[name_stream].extend(module_packages)
-                        self.repos.packages[package_name].extend(module_packages)
+                        self.repos.pkgs_from_modules[name_stream].extend(rpms + debug_rpms)
+                        self.repos.packages[package_name].extend(rpms)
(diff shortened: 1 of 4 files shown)


JayZ12138:
LGTM, just one thing. Now you are searching rpms for every package included in the artifacts, do you want to make it parallel?

ubipop/__init__.py
```diff
@@ -623,28 +628,44 @@ def get_packages_names_by_profiles(self, profiles, modules):
...
                     continue
                 if package_name and name != package_name:
                     continue
+                rpm_filename = rpm_without_epoch + '.rpm'
 
-                rpms.append(Package(name, rpm_without_epoch + '.rpm', is_modular=True))
-
-        return rpms
+                # Check existence of rpm in binary rpm repo
+                rpms = self.pulp.search_rpms(self.repos.in_repos.rpm, filename=rpm_filename)
```
danrodrig: I think that we can rework this to do a bulk query for the package list in artifacts. Repo ID is in package metadata, so we can split rpm and debug-rpm from there.

midnightercz: +1. I think for the future it make sense to search for more than one rpm at the time

rbikar: good points guys, I'll add it to my notes and file issue for optimizations.

rbikar:
I'm carrying on +1s as the last commit was solving conflict with master branch.

PR #95, opened 2019-05-14
Use threading.local to store session

### LINKED ISSUE #57: requests.Session should not be shared between threads
ubi-population-tool Pulp class is creating one requests.Session and allowing it to be used from multiple threads.

I'd like to suggest instead using a separate session per thread.  The reasons are:

- doubts whether requests.Session is really thread-safe - the documentation for requests doesn't elaborate on this and there's no end of resources such as https://toolbelt.readthedocs.io/en/latest/threading.html claiming it's only "mostly" thread-safe
- even if requests.Session itself is thread-safe, it's easy to accidentally use it in a way which is not thread-safe, for example mounting and unmounting adapters (note an adapter is being introduced by https://github.com/release-engineering/ubi-population-tool/pull/53)

Easiest way to do this is roughly to replace `self.session` with `self.local.session`, where `self.local` is a `threading.local`.

alexandrevicenzi:
With this the session won't be shared across
multiple threads.

This fix #57.

### DIFF
--- a/ubipop/_pulp_client.py
+++ b/ubipop/_pulp_client.py
@@ -48,15 +48,16 @@ class Pulp(object):
 
     def _make_session(self):
         adapter = PulpRetryAdapter()
-
-        self.local.session = requests.Session()
-        self.local.session.mount('http://', adapter)
-        self.local.session.mount('https://', adapter)
+        session = requests.Session()
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
 
         if len(self.auth) == 1:
-            self.local.session.cert = self.auth[0]
+            session.cert = self.auth[0]
         else:
-            self.local.session.auth = self.auth
+            session.auth = self.auth
+
+        self.local.session = session
 
     def do_request(self, req_type, url, data=None):
         if not hasattr(self.local, 'session'):
--- a/ubipop/_pulp_client.py
+++ b/ubipop/_pulp_client.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import threading
 import time
 
 from urllib3.util.retry import Retry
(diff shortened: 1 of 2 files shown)


ubipop/_pulp_client.py
```diff
@@ -42,30 +43,34 @@ def __init__(self, hostname, auth, insecure=False):
...
+        self.local = threading.local()
 
     def _make_session(self):
-        self.session = requests.Session()
-        self.adapter = PulpRetryAdapter()
-        self.session.mount('http://', self.adapter)
-        self.session.mount('https://', self.adapter)
+        adapter = PulpRetryAdapter()
+
+        self.local.session = requests.Session()
```
rohanpm: To me it would make more sense if self.local.session is assigned only once at the end.

The reason is that do_request and hence _make_session can be called concurrently from multiple threads.  Two threads can both see that self.local.session was missing, and start constructing it here.

It can happen for instance that one thread sets self.local.session = requests.Session(), mounts the 'http://' adapter; then another thread resets self.local.session again to a new instance.  The resulting session would be constructed with an odd and unpredictable sequence of calls, e.g. it could have one mount for 'http://', two mounts for 'https://' with two different adapters.

I can't point to any specific bug which can occur due to that, which is why I don't give -1, but it seems like it'd be more robust if self.local.session was only assigned once at the end - so at least you know that self.local.session is always either missing, or pointing at a fully configured requests.Session which was configured by a single thread with a known sequence of calls.

alexandrevicenzi:
> Code looks good, but I think tests should be improved. What we should test here is that if you create two threads then pulp client in thread1 shoulnd't be identical to pulp client in thread 2

Pulp client will be the same in both thread, this is the design on ubipop, that's why the session was shared. The only thing that should not be shared is the session.

ubipop/_pulp_client.py
alexandrevicenzi: If it's two different threads the session wont exists. This is not a concurrency problem because the session is per thread and not shared. The problem is if for some reason the same thread would execute this piece of code twice at the same time.

rohanpm: Yeah, of course.  The separate threads would each be assigning to their own self.local.session = requests.Session(), so my comment makes no sense.

midnightercz:
> Pulp client will be the same in both thread, this is the design on ubipop, that's why the session was shared. The only thing that should not be shared is the session.

Ok, sorry I chose wrong word. I meant session, not pulp client. So Basically, you can see it like this:

```python
import threading
import random

class Client1(object):
    def __init__(self):
        self.local = threading.local()
    def make_session(self):
        if not hasattr(self.local, "session"):
            self.local.session = random.randrange(1000) 
    def session(self):
        return self.local.session

def do_client_thing(c):
    c.make_session()
    print id(c.session())

c = Client1()
ts = [threading.Thread(target=do_client_thing, args=(c,)) for _ in range(5)]
[t.start() for t in ts]
[t.join() for t in ts]
```
And then to test it, it should return 5 different object ids. That's how you can test session isn't shared.

PR #101, opened 2019-05-22
Custom calculation of is_modular field

### LINKED ISSUE #94: Do not trust 'is_modular' field
**Description:**
Currently we trust **is_modular** field in rpm units. But this field is not really reliable.
It can only be trusted, if the value is True. 
If it's False it can be modular and also non-modular.

We should implement check for this field:
E.g. maybe something like: take rpm, check it's existence in any modulemd associated to repo and act accordingly. 

*Acceptance criteria:*
- Do not use value of **is_modular** field from pulp
- When adding rpms to output set from modules artifacts, set always is_modular to True
- When adding rpms to output set from binary rpm repo, do search for that rpm in associated modulemd units in input repo and set *is_modular* accordingly.

rbikar:
This change introduces custom calculation
of is_modular field of rpm unit. This field
is not taken from pulp unit metadata anymore because
it is unreliable.

is_modular field is true in case package comes from
artifacts of modulemd directly or is referenced
by any modulemd artifacts in input repository.

Fixes #94

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -128,6 +128,7 @@ def test_get_packages_from_module_by_name(mock_ubipop_runner):
     pkg = rpms[0]
     # filename is without  epoch
     assert pkg.filename == "postgresql-9.6.10-1.module+el8+2470+d1bafa0e.x86_64.rpm"
+    assert pkg.is_modular is True
 
 
 def test_get_packages_from_module(mock_ubipop_runner):
@@ -152,7 +153,7 @@ def test_get_packages_from_module(mock_ubipop_runner):
     rpms, debug_rpms = mock_ubipop_runner.get_packages_from_module(input_modules)
     assert len(rpms) == 2 # srpm is not included
     assert len(debug_rpms) == 0  # no debug rpm in this testcase
-
+    assert all([rpm.is_modular for rpm in rpms])
 
 def test_get_packages_from_module_debuginfo(mock_ubipop_runner):
     input_modules = [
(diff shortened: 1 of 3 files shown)


tests/test_ubipop.py
```diff
@@ -310,13 +311,23 @@ def _f(*args, **kwargs):
...
 
 
-def test_match_binary_rpms(mock_ubipop_runner):
-    package_name = 'foo-pkg'
-    mock_ubipop_runner.pulp.search_rpms.side_effect = _get_search_rpms_side_effect(package_name)
-    mock_ubipop_runner._match_binary_rpms() # pylint: disable=protected-access
+@pytest.mark.parametrize("pkg, modular", [("foo-no-match", False), ("foo-pkg", True)])
+def test_match_binary_rpms(mock_ubipop_runner, pkg, modular):
+    mock_ubipop_runner.pulp.search_modules.return_value = \
+        [get_test_mod(name="m1",
```
alexandrevicenzi: do like 
```py
foo = [
  'bar',
]
```

tests/test_ubipop.py
```diff
@@ -972,3 +983,23 @@ def test_exclude_blacklisted_packages(mock_ubipop_runner):
     # no blacklisting from pkgs from mds
     assert len(mock_ubipop_runner.repos.pkgs_from_modules) == 1
     assert len(mock_ubipop_runner.repos.debug_rpms) == 0
+
+
+def test_get_pkgs_from_all_modules(mock_ubipop_runner):
+    mock_ubipop_runner.pulp.search_modules.return_value = \
```
alexandrevicenzi: same here

ubipop/__init__.py
```diff
@@ -234,11 +234,25 @@ def _match_module_defaults(self):
             if module_defaults:
                 self.repos.module_defaults[fts[ft]].extend(module_defaults)
 
+    def _get_pkgs_from_all_modules(self):
+        modules = self.pulp.search_modules(self.repos.in_repos.rpm)
+        pkgs = set()
+        regex = r'\d+:'
```
alexandrevicenzi: regex seems kinda useless, not reused. Maybe just `compile(r'\d+:')`.

PR #103, opened 2019-05-24
Don't blacklist source RPMs referenced by modules

### LINKED ISSUE #97: Do not use blacklist for source RPMs that are referenced by modular RPMs
**Description:**
At the moment when creating output set for source RPMs we blacklist all source RPMs that match with ones in blacklist. This seems not to be a wanted behavior.
Since don't blacklist modular RPMs we shouldn't blacklist any source RPMs that is referenced from modular RPM.

**Acceptace crit.:**
- Do not blacklist source RPMs that are referenced by modular RPMs

**Note:**
 - This change will rely on #94  (is_modular reliability)

nathanegillett:
Since there's no blacklist for modules, "excluded" source RPMs
referenced by modular RPMs are no longer added to the blacklist.

Fixes #97

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -350,7 +350,8 @@ class UbiPopulateRunner(object):
             list(chain.from_iterable(self.repos.source_rpms.values())))
 
         for pkg in blacklisted_srpms:
-            self.repos.source_rpms.pop(pkg.name, None)
+            if not pkg.is_modular:
+                self.repos.source_rpms.pop(pkg.name, None)
 
     def _determine_pulp_actions(self, units, current, diff_f):
         expected = list(chain.from_iterable(units.values()))


ubipop/__init__.py
```diff
@@ -575,6 +575,7 @@ def get_blacklisted_packages(self, package_list):
         Finds blacklisted packages in output sets
         """
         blacklisted_pkgs = []
+        modular_rpms = [m.name for m in self.ubiconfig.modules.whitelist]
```
rbikar: This doesn't seem right.
self.ubiconfig.modules.whitelist gives you names of modules, not modular rpm name

nathanegillett: Hmm. self.ubiconfig.modules.whitelist[0].name is the module name, not the modular rpm name? If so, how does one get the rpm names?

rbikar: - First thing about pkgs names - in ubi config there is a whitelist and a blacklist for pkgs in form of simple regex. See usage at https://github.com/release-engineering/ubi-population-tool/blob/2081f6846c6ebdf72036ec2c3d25b9d8afe4d225/ubipop/__init__.py#L257

- Next thing that I already mentioned in different comment  - imho this change should happen in https://github.com/release-engineering/ubi-population-tool/blob/2081f6846c6ebdf72036ec2c3d25b9d8afe4d225/ubipop/__init__.py#L338
and since #101 was merged, you can rely on **is_modular** attribute of Package objects.

nathanegillett: > * Next thing that I already mentioned in different comment  - imho this change should happen in https://github.com/release-engineering/ubi-population-tool/blob/2081f6846c6ebdf72036ec2c3d25b9d8afe4d225/ubipop/__init__.py#L338

Done

>   and since #101 was merged, you can rely on **is_modular** attribute of Package objects.

Very handy. Thank you.

rbikar:
I also noticed that you created branch in upstream repo, I guess it's not something really blocking but I think common practice is that everyone works on his forked repo and branch and not on upstream repo  and creates PR from forked repo/branch to upstream/master.
You may want to check https://help.github.com/en/articles/fork-a-repo and https://help.github.com/en/articles/syncing-a-fork

nathanegillett:
> Looks good, but I would like to see explicit test case for modular rpm/srpm. Would you mind adding one? (Or I guess just updating existing testcase also with modular rpm)
> I know that code is now covered but I think it'll be better to have unit test for that case.

Been working on this but my knowledge of the content and code is still developing.

nathanegillett:
> I also noticed that you created branch in upstream repo, I guess it's not something really blocking but I think common practice is that everyone works on his forked repo and branch and not on upstream repo and creates PR from forked repo/branch to upstream/master.
> You may want to check https://help.github.com/en/articles/fork-a-repo and https://help.github.com/en/articles/syncing-a-fork

I'll give those a look. Thank you!

PR #106, opened 2019-05-29
Don't blacklist source RPMs referenced by modules

### LINKED ISSUE #97: Do not use blacklist for source RPMs that are referenced by modular RPMs

nathanegillett:
Since there's no blacklist for modules, excluded source RPMs referenced
by modular RPMs are no longer added to the blacklist. Fixes #97.

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -625,26 +625,34 @@ def test_ubipopulate_load_all_ubiconfig(mocked_ubiconfig_load_all):
 
 
 def test_create_srpms_output_set(mock_ubipop_runner):
-    expected_src_rpm_filename = "tomcatjss-7.3.6-1.el8+1944+b6c8e16f.src.rpm"
     mock_ubipop_runner.repos.packages['foo'] = [
         get_test_pkg(
             name="tomcatjss",
             filename="tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm",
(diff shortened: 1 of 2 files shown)


tests/test_ubipop.py
```diff
@@ -642,9 +648,8 @@ def test_create_srpms_output_set(mock_ubipop_runner):
     mock_ubipop_runner._create_srpms_output_set() # pylint: disable=protected-access
 
     out_srpms = mock_ubipop_runner.repos.source_rpms
-    assert len(out_srpms) == 1
-    assert out_srpms['tomcatjss'][0].name == "tomcatjss"
-    assert out_srpms['tomcatjss'][0].filename == expected_src_rpm_filename
+    assert len(out_srpms) == 2
+    assert "tomcatjss" and "foo-pkg" in out_srpms
```
rbikar: No really blocking but consider checking that  filenames are correct.

nathanegillett: Done.

PR #110, opened 2019-06-06
Replace cmp_version with rpm python bindings

### LINKED ISSUE #104: cmp_version is unable to correctly sort rpm with uneven release parts
*Description:*
There is an issue with comparison of rpm version in cmp_version module.
For details see https://github.com/kata198/cmp_version/issues/1 

*Note:*
I created PR https://github.com/kata198/cmp_version/pull/2  that should fix this issue
but maintainer of that repo hasn't responded yet. 
Until upstream cmp_version is released with fix,  would it be ok to incorporate source code of cmp_version to ubipop code? Or can you think of any better solution? 
@danrodrig   @rohanpm  @JayZ12138

rbikar:
This change replaces buggy cmp_version package
with python bindings of official rpm project.

This should ensure that ubipop will sort rpms by correctly implemented and maitained vercmp algorithm.

Fixes #104

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 requests
 more-executors
 ubi-config
-cmp_version
+rpm-py-installer
--- a/ubipop/_pulp_client.py
+++ b/ubipop/_pulp_client.py
@@ -4,6 +4,7 @@ import threading
 import time
 
 from urllib3.util.retry import Retry
+from ubipop._utils import split_filename
 
 try:
     from urllib.parse import urljoin
@@ -11,8 +12,8 @@ except ImportError:
     from urlparse import urljoin
 
 import requests
+from rpm import labelCompare as label_compare  # pylint: disable=no-name-in-module
 
-from cmp_version import cmp_version
 
 _LOG = logging.getLogger("ubipop")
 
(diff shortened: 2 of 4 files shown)


ubipop/_pulp_client.py
```diff
@@ -4,15 +4,16 @@
...
 from urllib3.util.retry import Retry
+from ubipop._utils import split_filename
 
 try:
     from urllib.parse import urljoin
 except ImportError:
     from urlparse import urljoin
 
 import requests
+from rpm import labelCompare as label_compare  # pylint: disable=no-name-in-module
```
alexandrevicenzi: Instead of adding an extra dep, would not be easy to just borrow this function?

JayZ12138: yeah, sounds like a plan.

rohanpm: But it's implemented in C, isn't it?  I would rather keep our stuff pure-python as much as we can.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 requests
 more-executors
 ubi-config
-cmp_version
+rpm-py-installer
```
rohanpm: I think this might cause some issues for the RPM packages of this project.  Is this going to result in the egg always requiring some "rpm-py-installer" distribution to be present, while it's actually nonsense to set that as Requires in an RPM? 

In general though I think it's OK for what we put in these repos to be somewhat ignorant of RPM packaging complexities.  So this comment is more like a heads up that this might cause us to run into issues and we may need patches in dist-git, or may need to revisit this later.

ubipop/_pulp_client.py
rbikar: Yep, it's in C, I also wouldn't like to have extra C code in our stuff.

alexandrevicenzi: Oh, if it's in C makes sense.

PR #114, opened 2019-06-20
Conditional creation of install_requires of rpm distribution

rbikar:
This commit adds conditional creation of install_requires list
in setup.py for rpm distribution.
If rpm-py-installer was added there unconditionally,
rpm of ubipop couldn't be installed on some envs.
If implemented this way there is no need to change
anything in .spec file for rpm builds.

This fix is based on rpm-py-installer docs:
See Q6 in FAQ ->
https://github.com/junaruga/rpm-py-installer/blob/master/docs/users_guide.md

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 requests
 more-executors
 ubi-config
-rpm-py-installer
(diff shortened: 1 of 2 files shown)


rbikar:
This requires #115 for passing travis CI.

rbikar:
Rebased on #115,  I'll carry over +1s

rbikar:
ok test

PR #115, opened 2019-06-20
Explicit dist: trusty in travis.yml for py26 env

rbikar:
This change explicitly sets dist in travis
to 'trusty' for py26 env.
The reason is that 'xenial' was made default dist on travis
recently and it does not support python 2.6 anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-dist: trusty
 sudo: false
 language: python
 before_install:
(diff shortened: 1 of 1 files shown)


rohanpm:
ok test

.travis.yml
```diff
@@ -1,3 +1,4 @@
+dist: trusty
```
alexandrevicenzi: Can you move dist to the matrix? Otherwise will use trusty for every Python version.

rbikar: yep -> done

rbikar:
ok test

Issue #74, opened 2019-05-01
Duplicated source rpm association


emilyzheng:
When associating the same package in different arches (which have the same source rpm) to ubi repo, their source rpm will be associated multiple times.  E.g.:
```
2019-04-30 22:04:35 +0000 [INFO    ] Would associate glibc-2.17-195.el7.x86_64.rpm from rhel-7-server-rpms__7_DOT_1__x86_64 to ubi-7-server-rpms__7_DOT_1__x86_64
2019-04-30 22:04:35 +0000 [INFO    ] Would associate glibc-2.17-195.el7.i686.rpm from rhel-7-server-rpms__7_DOT_1__x86_64 to ubi-7-server-rpms__7_DOT_1__x86_64
2019-04-30 22:04:35 +0000 [INFO    ] Would associate glibc-2.17-195.el7.src.rpm from rhel-7-server-source-rpms__7_DOT_1__x86_64 to ubi-7-server-source-rpms__7_DOT_1__x86_64
2019-04-30 22:04:35 +0000 [INFO    ] Would associate glibc-2.17-195.el7.src.rpm from rhel-7-server-source-rpms__7_DOT_1__x86_64 to ubi-7-server-source-rpms__7_DOT_1__x86_64 
```
This should be avoided to improve efficiency.

danrodrig:
This was fixed in #117.

PR #117, opened 2019-07-11
Remove duplicate associations for srpm packages (#74)

midnightercz:
When making list of srpm packages, there are duplicates in output
list due fact that many rpms can share one srpm. That leads to
unnecessary duplicated associations.
That was fixed in this commit

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -376,6 +376,19 @@ class UbiPopulateRunner(object):
     def _get_pulp_actions_pkgs(self, pkgs, current):
         return self._determine_pulp_actions(pkgs, current, self._diff_packages_by_filename)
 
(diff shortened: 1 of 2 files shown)


danrodrig:
ok test

ubipop/__init__.py
```diff
@@ -376,6 +376,19 @@ def _get_pulp_actions_md_defaults(self, module_defaults, current):
     def _get_pulp_actions_pkgs(self, pkgs, current):
         return self._determine_pulp_actions(pkgs, current, self._diff_packages_by_filename)
 
+    def _get_pulp_actions_src_pkgs(self, pkgs, current):
```
JayZ12138: I might be a fool..., but some explanation would be nice. I could hardly understand the code until I checked the test case.

midnightercz:
ok test

PR #118, opened 2019-07-12
When insecure=True disable SSL warnings logging

midnightercz:
When insecure=True is pass to Pulp client instance it's expected that
ssl certificate can't be verified, therefore it's not neccessary to
keep SSL warning messages enabled.


danrodrig:
ok test

PR #123, opened 2019-07-22
Avoid duplicates for packages present in modules and white list

### LINKED ISSUE #105: keep_n_latest_packages() may add duplicated packages to output set
**Description:**
keep_n_latest_packages()  method can add duplicates to output set of rpms in case they matched  by whitelist and also when the same package comes from module's artifacts.
This should be avoided so we don't have duplicated associations.

What most probably happens is that there are two packages with the same name and filename but is_modular attr. differs. Current implementation will add both of those pkgs to output set and they are both associated.
**Acceptance criteria:**
* Output sets of binary and debug rpms should contain only unique rpms
* There is no duplicated associated of binary and debug rpms

**Note:**
* There is also issue  #52 that was partially fixed but we still have duplicated associations for source rpms but the reason is not only is_modular attr. but there are also multilib pkgs that the same sourcerpm (x.x86_64.rpm and x.i686.rpm have the same SRPM)

danrodrig:
In keep_n_latest_packages(), keep track of filenames for packages added from modules and skip them while processing packages from white list.

Add test to ensure that only package from modules is kept.

Refactor keep_n_latest_packages():
- Move filter per arches of non-modular to the main packages loop
- Use collections.deque as queue to keep n package versions per arch
- Build set of package filenames from modules for reference checks outside the main packages loop.
- Extend doc string

Fixes #105

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -569,6 +569,44 @@ def test_keep_n_newest_packages_with_referenced_pkg_in_module(mock_ubipop_runner
     assert "7.3.7" in packages[0].filename
 
 
+def test_keep_n_latest_packages_no_dupes_from_modular(mock_ubipop_runner):
+    """Ensure that modular packages referenced also in the package
+     white list are included only once in the output list.
+
+    Notice that this situation was observed only on integration tests
+    due to the test data used.
+    In practice, packages in modules have "module" as part of the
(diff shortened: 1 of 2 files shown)


danrodrig:
ok test

danrodrig:
Decreased coverage is due to removed lines.

danrodrig:
ok test

PR #124, opened 2019-07-25
Fix unique filter for srpms

midnightercz:
release and version subkeys were added to srpm unique filter to work
correctly in situations where srpms with same name but different
versions or releases needs to be associated.


midnightercz:
ok test

PR #126, opened 2019-08-26
Enable population by content set(s) or repo ID(s)

### LINKED ISSUE #121: Ubipop is able to run only for specific content set or repository
#### Summary
The goal of this issue is to add support for running only for specific content sets and repositories in ubipop. Now it is able to run per config file granularity but we need to also run it for smallest possible units (content set and repo)

#### Issue Type
- Feature Request

#### Desired behavior
- ubipop accepts specific content set label or repo id (1 or more) for content calculation
- ubipop is able to to filter loaded ubi config by content set label for input content set or given repo id.
- public methods created are properly documented

nathanegillett:
UbiPopulate now accepts specific content set lables and/or repository
IDs for content calculation.

Fixes #121

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -80,19 +80,18 @@ class UbiRepoSet(object):
 
 class UbiPopulate(object):
     def __init__(self, pulp_hostname, pulp_auth, dry_run, ubiconfig_filename_list=None,
-                 ubiconfig_dir_or_url=None, content_sets=None, repo_ids=None, insecure=False,
-                 workers_count=4, output_repos=None):
+                 ubiconfig_dir_or_url=None, insecure=False, workers_count=4, output_repos=None,
+                 **kwargs):
 
         self.pulp = Pulp(pulp_hostname, pulp_auth, insecure)
         self.dry_run = dry_run
         self.output_repos = output_repos
         self._executor = Executors.thread_pool(max_workers=workers_count).with_retry()
-        self.ubiconfig_list = self._load_ubiconfig(filenames=ubiconfig_filename_list,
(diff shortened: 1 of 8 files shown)


ubipop/__init__.py
```diff
@@ -80,24 +80,53 @@ def _ensure_repos_existence(self):
 
 class UbiPopulate(object):
     def __init__(self, pulp_hostname, pulp_auth, dry_run, ubiconfig_filename_list=None,
-                 ubiconfig_dir_or_url=None, insecure=False, workers_count=4,
-                 output_repos=None):
+                 ubiconfig_dir_or_url=None, content_sets=None, repo_ids=None, insecure=False,
```
alexandrevicenzi: There are too many parameters I believe

ubipop/cli.py
```diff
@@ -15,7 +15,11 @@ def parse_args(args):
     parser = argparse.ArgumentParser()
     parser.add_argument('input', action="store", nargs='*',
                         help="path to ubi config file")
-    parser.add_argument('--conf-src', action="store",
+    parser.add_argument('--content-sets', action="store", nargs='+', type=str, required=False,
+                        help="one or more content set labels")
```
midnightercz: I can imagine better help/explanation for the attribute

ubipop/__init__.py
midnightercz: also if you remember presentation about API design, this breaks compatibility.

rohanpm: This isn't public API though... officially, there *is* no public API for this library yet.

nathanegillett:
run tests

PR #127, opened 2019-08-29
Only populate repos marked for population

### LINKED ISSUE #92: Skip populating _DOT_ repos with equal or older than current release version
Currently, the UBI population does various work on DOT repos, such as 

> 2019-05-07 18:54:42 +0000 [INFO    ] Publishing ubi-7-server-rpms__7_DOT_1__x86_64 in yum_distributor
> 2019-05-07 18:54:42 +0000 [INFO    ] Publishing ubi-7-server-debug-rpms__7_DOT_1__x86_64 in yum_distributor
> 2019-05-07 18:54:42 +0000 [INFO    ] Publishing ubi-7-server-source-rpms__7_DOT_1__x86_64 in yum_distributor
> 2019-05-07 18:57:40 +0000 [INFO    ] Publishing ubi-7-server-rpms__7_DOT_0__x86_64 in yum_distributor
> 2019-05-07 18:57:40 +0000 [INFO    ] Publishing ubi-7-server-debug-rpms__7_DOT_0__x86_64 in yum_distributor
> 2019-05-07 18:57:40 +0000 [INFO    ] Publishing ubi-7-server-source-rpms__7_DOT_0__x86_64 in yum_distributor

However, we are not publishing the older versions to the public and the only point of the DOT repos for UBI is to make it possible to prepare pre-release content. The UBI population should skip these older repos.

nathanegillett:
Using the new repo note, ubi_population, this commit ensures repo sets
that would contain repos not intended for population are skipped.

Fixes #92

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -16,15 +16,15 @@ from ubipop._pulp_client import Repo
 
 def test_raise_not_implemented_pulp_action():
     units = ["unit1", "unit2"]
-    repo = Repo("test", "test-rpms", "1", None, "2", None)
+    repo = Repo("test", "1", "test-rpms", "2", None, None)
     action = PulpAction(units, repo)
     pytest.raises(NotImplementedError, action.get_action, None)
 
 
 def test_raise_not_implemented_associate_action():
     units = ["unit1", "unit2"]
-    repo = Repo("test", "test-rpms", "1", None, "2", None)
-    src_repo = Repo("test", "test-rpms", "1", None, "2", None)
+    repo = Repo("test", "1", "test-rpms", "2", None, None)
+    src_repo = Repo("test", "1", "test-rpms", "2", None, None)
     action = AssociateAction(units, repo, src_repo)
     pytest.raises(NotImplementedError, action.get_action, None)
 
@@ -36,8 +36,8 @@ def test_raise_not_implemented_associate_action():
 ])
 def test_get_action_associate(klass, method):
     units = ["unit1", "unit2"]
-    dst_repo = Repo("test_dst", "test_dst-rpms", "1", None, "2", None)
-    src_repo = Repo("test_src", "test_src-rpms", "1", None, "2", None)
+    dst_repo = Repo("test_dst", "1", "test_dst-rpms", "2", None, None)
+    src_repo = Repo("test_src", "1", "test_src-rpms", "2", None, None)
     action = klass(units, dst_repo, src_repo)
     associate_action, src_repo_current, dst_repo_current, current_units = \
         action.get_action(MagicMock())
@@ -55,7 +55,7 @@ def test_get_action_associate(klass, method):
 ])
 def test_get_action_unassociate(klass, method):
     units = ["unit1", "unit2"]
-    dst_repo = Repo("test_dst", "test_dst-rpms", "1", None, "2", None)
+    dst_repo = Repo("test_dst", "1", "test_dst-rpms", "2", None, None)
     action = klass(units, dst_repo)
     associate_action, dst_repo_current, current_units = action.get_action(MagicMock())
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -16,15 +16,15 @@ from ubipop._pulp_client import Repo
 
 def test_raise_not_implemented_pulp_action():
     units = ["unit1", "unit2"]
-    repo = Repo("test", "test-rpms", "1", "2", None)
+    repo = Repo("test", "test-rpms", "1", None, "2", None)
     action = PulpAction(units, repo)
     pytest.raises(NotImplementedError, action.get_action, None)
 
 
 def test_raise_not_implemented_associate_action():
     units = ["unit1", "unit2"]
-    repo = Repo("test", "test-rpms", "1", "2", None)
-    src_repo = Repo("test", "test-rpms", "1", "2", None)
+    repo = Repo("test", "test-rpms", "1", None, "2", None)
+    src_repo = Repo("test", "test-rpms", "1", None, "2", None)
     action = AssociateAction(units, repo, src_repo)
     pytest.raises(NotImplementedError, action.get_action, None)
 
@@ -36,8 +36,8 @@ def test_raise_not_implemented_associate_action():
 ])
 def test_get_action_associate(klass, method):
     units = ["unit1", "unit2"]
-    dst_repo = Repo("test_dst", "test_dst-rpms", "1", "2", None)
-    src_repo = Repo("test_src", "test_src-rpms", "1", "2", None)
+    dst_repo = Repo("test_dst", "test_dst-rpms", "1", None, "2", None)
+    src_repo = Repo("test_src", "test_src-rpms", "1", None, "2", None)
     action = klass(units, dst_repo, src_repo)
     associate_action, src_repo_current, dst_repo_current, current_units = \
         action.get_action(MagicMock())
@@ -55,7 +55,7 @@ def test_get_action_associate(klass, method):
 ])
 def test_get_action_unassociate(klass, method):
     units = ["unit1", "unit2"]
-    dst_repo = Repo("test_dst", "test_dst-rpms", "1", "2", None)
+    dst_repo = Repo("test_dst", "test_dst-rpms", "1", None, "2", None)
     action = klass(units, dst_repo)
     associate_action, dst_repo_current, current_units = action.get_action(MagicMock())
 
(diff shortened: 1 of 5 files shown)


nathanegillett:
run tests

nathanegillett:
run tests

ubipop/__init__.py
```diff
@@ -148,18 +149,27 @@ def _get_input_and_output_repo_pairs(self, ubiconfig_item):
...
+            in_debug_info = self._get_repo_counterpart(input_repo, in_debug_repos_ft.result())
+
+            out_rpm = self._get_repo_counterpart(input_repo, out_repos_ft.result())
+            out_source = self._get_repo_counterpart(input_repo, out_source_repos_ft.result())
+            out_debug_info = self._get_repo_counterpart(input_repo, out_debug_repos_ft.result())
+
+            # identify any ubi repos not marked for population
+            no_populate = []
+            for repo in (out_rpm, out_source, out_debug_info):
+                if repo.ubi_population is False:
```
alexandrevicenzi: I just wonder that should happen if ubi_population is None, 0, an empty string or everything else that could fall under the previous `if not ubi_population`.

You enforce that the value is False, I would change and check if the value is True, everything else is a misleading value.

alexandrevicenzi: I also think that list comprehension is still better looking than this for loop.

nathanegillett: It is possible for ubi_population to be None (repos that haven't yet had the note applied) which is why I changed this code to check specifically for False.

As for reversing the check, we would have to ensure that the value is True for all three output repos. Doing so may over-complicate this filter but I can try to come up with something.

nathanegillett:
run tests

ubipop/__init__.py
```diff
@@ -148,19 +149,23 @@ def _get_input_and_output_repo_pairs(self, ubiconfig_item):
...
+            in_source = self._get_repo_counterpart(input_repo, in_source_repos_ft.result())
+            in_debug_info = self._get_repo_counterpart(input_repo, in_debug_repos_ft.result())
+
+            out_rpm = self._get_repo_counterpart(input_repo, out_repos_ft.result())
+            out_source = self._get_repo_counterpart(input_repo, out_source_repos_ft.result())
+            out_debug_info = self._get_repo_counterpart(input_repo, out_debug_repos_ft.result())
+
+            # skip repo set if not all output repos are intended for population
+            can_pop = [r for r in (out_rpm, out_source, out_debug_info) if r.ubi_population is True]
+            if len(can_pop) == 3:
```
alexandrevicenzi: Replace this if with `all`.

```py
can_pop = all([r.ubi_population for r in (out_rpm, out_source, out_debug_info)])

if can_pop:
  # code
```

nathanegillett: All three repos need to pass to create the repo set.

nathanegillett: This doesn't work anyway because not all repos have ubi_population yet.

nathanegillett:
run tests

nathanegillett:
rebased on #126

nathanegillett:
run tests

ubipop/__init__.py
alexandrevicenzi: why does not work? change to `all([r.ubi_population is True for r in (out_rpm, out_source, out_debug_info)])`, it should be the same, I just think that this if equals 3 is not that readable, kinda magic number there.

ubipop/_pulp_client.py
```diff
@@ -92,6 +92,7 @@ def _search_repo(self, criteria):
                 arch=notes['arch'],
                 content_set=notes['content_set'],
                 platform_full_version=notes['platform_full_version'],
+                ubi_population=notes.get('ubi_population', None),  # ubi repos only
```
midnightercz: why explicitly None?

ubipop/_pulp_client.py
```diff
@@ -306,11 +307,13 @@ def publish_repo(self, repo):
 
 
 class Repo(object):
-    def __init__(self, repo_id, arch, content_set, platform_full_version, dist_ids_type_ids):
+    def __init__(self, repo_id, arch, content_set, platform_full_version, ubi_population,
```
midnightercz: this breaks API, 'ubi_population' should be added to the end of method

ubipop/_pulp_client.py
nathanegillett: Accuracy? I suppose we could default to True since we're allowing repos without ubi_population note.

ubipop/_pulp_client.py
nathanegillett: I don't think this is a public API since the module is led by an underscore and all the internal references were updated, so nothing should be broken. I added the content_set parameter similarly.

ubipop/__init__.py
nathanegillett: It doesn't work because not all repos have ubi_population note and there doesn't seem to be incentive to set this note to True so most don't have it at all. The idea, I think, is to let cdn-utils set this note on all repos over time as there republished -- no "big bang" is planned to do this for all repos so that they have these new notes.

However, all repos that aren't intended for population seem to already have the ubi_population set to False because incentive to do so exists.

@midnightercz brought up the fact that I have the default for this field set to None. What I could do is default to True for those repos which don't have ubi_population. Then, I could do as  you suggest.

nathanegillett:
run tests

ubipop/_pulp_client.py
```diff
@@ -306,12 +310,16 @@ def publish_repo(self, repo):
...
-    def __init__(self, repo_id, arch, content_set, platform_full_version, dist_ids_type_ids):
+    def __init__(self, repo_id, arch, content_set, platform_full_version, dist_ids_type_ids,
+                 ubi_population):
         self.repo_id = repo_id
         self.arch = arch
         self.content_set = content_set
         self.platform_full_version = platform_full_version
         self.distributors_ids_type_ids_tuples = dist_ids_type_ids
+        # Only assign ubi_population for UBI repos
+        if ubi_population is not None:
```
danrodrig: I don't think it's a good idea to make conditional the presence of the ubi_population property in Repo objects.
As ubi_population is optional in repo notes, make it optional as a keyword argument, default to None, and use it to initialize self.ubi_population.
And would be nice to document the argument. I don't think we have any agreement on how to document types. I like to use mypy syntax, "Optional[bool]", in the doc string.

ubipop/__init__.py
```diff
@@ -180,18 +181,21 @@ def _get_input_and_output_repo_pairs(self, ubiconfig_item):
...
-            source = self._get_repo_counterpart(input_repo, out_source_repos_ft.result())
-            debug_info = self._get_repo_counterpart(input_repo, out_debug_repos_ft.result())
+            out_rpm = self._get_repo_counterpart(input_repo, out_repos_ft.result())
+            out_source = self._get_repo_counterpart(input_repo, out_source_repos_ft.result())
+            out_debug_info = self._get_repo_counterpart(input_repo, out_debug_repos_ft.result())
 
-            ubi_repo_set = RepoSet(rpm, source, debug_info)
+            # Skip repos sets containing output repos which should not be populated
+            if not all([r.ubi_population is True for r in (out_rpm, out_source, out_debug_info)]):
+                _LOG.debug("Skipping repo set for %s: should not be populated", in_rpm.repo_id)
```
danrodrig: I think it's better to be explicit in the logs about what repos are not being populated. Can you log the 3 repos?

nathanegillett:
run tests

tests/test_utils.py
```diff
@@ -16,15 +16,15 @@
 
 def test_raise_not_implemented_pulp_action():
     units = ["unit1", "unit2"]
-    repo = Repo("test", "test-rpms", "1", "2", None)
+    repo = Repo("test", "test-rpms", "1", None, "2", None)
```
danrodrig: These arguments don't look right.

`Repo(repo_id, arch, content_set, platform_full_version, dist_ids_type_ids, ubi_population)`

 dist_ids_type_ids is a list of tuples. So this should be something like:

`repo = Repo("test", "test-rpms", "1", "2", [], None)`

Or at least, keep that argument as None.

nathanegillett: Yes, thank you. The evolution of Repo's parameters hasn't been smooth. I'll make sure all Repo objects are correct.

nathanegillett:
run tests

nathanegillett:
> Only one thing, that I missed in previous reviews. You are not testing that UBI repositories without ubi_population note get populated. Also, would be better to have separate test or test for the ubi_population behavior.

This is actually already tested in test_pulp.py by test_search_repo_by_cs. We assert that ubi_population is True when no ubi_population repo note existed.

danrodrig:
> > Only one thing, that I missed in previous reviews. You are not testing that UBI repositories without ubi_population note get populated. Also, would be better to have separate test or test for the ubi_population behavior.
> 
> This is actually already tested in test_pulp.py by test_search_repo_by_cs. We assert that ubi_population is True when no ubi_population repo note existed.

I see, I overlooked that. Thanks for pointing it out.

PR #128, opened 2019-09-06
Add API for accessing expected Pulp actions

### LINKED ISSUE #119: Public interface for getting expected actions on pulp
#### Summary
Ubipop should provide public method for getting expected unassociations and associations that should happen on pulp for ubi repos. Information returned will be used in pub (or other clients) which will run these un/associations and save all received items to push item list.

Input of new method should be content set labels or repository ids.

Output - expected actions on repositories for rpms, srpms, modulemds and modulemd_defaults.

#### Issue Type
- Feature Request

#### Current behavior
There is no way how to get expected pulp actions out of ubipop now.

#### Desired behavior
- ubipop provides public method for getting expected actions on pulp
- ubi won't execute these actions but it'll leave them for caller

#### Use case
Client application needs to know what unit should be un/associated, so caller can execute it.

#### Additional Information
In ubipop there are already methods for calculation of expected pulp actions but it's not exposed publicly and it's implemented inside ubi repos population process.

nathanegillett:
This commit adds a new API which generates a list of the associations
and unassociations expected to be performed by Pulp for each repo in
each UBI configuration.

Fixes #119

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 import pytest
 import ubiconfig
 
-from mock import MagicMock, patch, call
+from mock import MagicMock, patch, call, ANY
 from more_executors import Executors
 from ubipop import UbiPopulateRunner, UbiRepoSet, RepoSet, UbiPopulate
 from ubipop._pulp_client import Module, ModuleDefaults, Package, Repo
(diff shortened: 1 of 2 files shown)


nathanegillett:
run tests

nathanegillett:
> I think that this way of implementation would be very difficult to use in client apps.
Okay, I can encapsulate associations and unassociations and leave out mdd_*.

> As this is 1st public API for ubipop, what I miss here is update of readme, docs with examples,
> examples dir with simple example of usage of this API.
> [...] What I mean is that UbiPopulate class will become public and there are no docs how to use it.
> Also classes that are not intended for public usage should be at least prefixed with underscore.
So, making existing classes "private" and writing all documentation for this UbiPopulate is now in scope of this issue, or just the new pulp actions API?

rbikar:
@nathanegillett 
Yeah, you're right docs UbiPopulate and other related things are not specified in this issue. So please file a new issue for this.

PR #129, opened 2019-09-12
Add API for accessing expected Pulp actions

### LINKED ISSUE #119: Public interface for getting expected actions on pulp

nathanegillett:
This commit adds a new API which generates a list of the associations
and unassociations expected to be performed by Pulp for the given
configuration.

Fixes #119

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -27,6 +27,7 @@ class RepoMissing(Exception):
 
 
 RepoSet = namedtuple('RepoSet', ['rpm', 'source', 'debug'])
+ExpectedPulpActions = namedtuple('ExpectedPulpActions', ['associations', 'unassociations'])
 
 
 class UbiRepoSet(object):
@@ -132,6 +133,43 @@ class UbiPopulate(object):
 
         return filtered_conf_list
 
+    def expected_pulp_actions(self, content_sets=None, repo_ids=None):
+        """
+        Returns a named tuple, ExpectedPulpActions, which contains a list of
+        associations and a list of unassociations expected to be performed by
+        Pulp for the loaded UBI configurations, optionally filtered by content
+        sets or repo IDs.
+
+        :param  content_sets:           list of content set labels by which to
+                                        filter UBI configurations
+
+        :param  repo_ids:               list of repos IDs by which to filter
+                                        UBI configurations
+
+        :return expected_pulp_actions:  namedtuple containing associations and
+                                        unassociations expected to be performed
+                                        by Pulp
+        """
+
+        expected_pulp_actions = ExpectedPulpActions([], [])
+
+        for config in self._filter_ubi_conf_list(self.ubiconfig_list, content_sets, repo_ids):
+            try:
+                repo_pairs = self._get_input_and_output_repo_pairs(config)
+            except RepoMissing:
+                _LOG.warning("Skipping current content triplet, some repos are missing")
+                continue
+
+            for repo_set in repo_pairs:
+                associations, unassociations = UbiPopulateRunner(
+                    self.pulp, repo_set, config, self.dry_run, self._executor
+                ).export_pulp_actions()
+
+                expected_pulp_actions.associations.extend(associations)
+                expected_pulp_actions.unassociations.extend(unassociations)
+
+        return expected_pulp_actions
+
     def populate_ubi_repos(self):
         out_repos = set()
 
@@ -218,6 +256,11 @@ class UbiPopulateRunner(object):
         self.ubiconfig = ubiconfig_item
         self.dry_run = dry_run
         self._executor = executor
+
+        self._associations = None
+        self._unassociations = None
+        self._mdd_association = None
+        self._mdd_unassociation = None
 
     def _match_modules(self):
         # Add matching modules
@@ -435,16 +478,19 @@ class UbiPopulateRunner(object):
 
         return self._determine_pulp_actions(src_pkgs, current, self._diff_packages_by_filename)
 
-    def _get_pulp_actions(self, current_modules_ft, current_module_defaults_ft, current_rpms_ft,
-                          current_srpms_ft, current_debug_rpms_ft):
+    def _set_pulp_actions(self):
         """
         Determines expected pulp actions by comparing current content of output repos and
-        expected content.
+        expected content. Stores expected pulp actions as UbiPopulateRunner attributes.
 
         Content that needs association: unit is in expected but not in current
         Content that needs unassociation: unit is in current but not in expected
         No action: unit is in current and in expected
         """
+
+        current_modules_ft, current_module_defaults_ft, current_rpms_ft, current_srpms_ft, \
+            current_debug_rpms_ft = self._get_current_content()
+
         modules_assoc, modules_unassoc = self._get_pulp_actions_mds(self.repos.modules,
                                                                     current_modules_ft.result())
         md_defaults_assoc, md_defaults_unassoc = \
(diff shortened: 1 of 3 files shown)


nathanegillett:
Ah, I forgot documentation/example.

alexandrevicenzi:
run tests

nathanegillett:
@rbikar & everyone else, I won't be adding docs for this, as I feel it would require the creation of docs for all of UbiPopulate. So, I think this can be included in the "make public" issue that I've yet to create.

ubipop/__init__.py
```diff
@@ -497,6 +517,28 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
 
         return diff
 
+    def expected_pulp_actions(self):
+        actions = ExpectedPulpActions([], [])
+
+        Association = namedtuple('Association', ['unit_id', 'type', 'src_repo', 'dest_repo'])
+        Unassociation = namedtuple('Unassociation', ['unit_id', 'type', 'dest_repo'])
+
+        associations, unassociations, _, _ = self._get_pulp_actions(*self._get_current_content())
```
rbikar: I think there was some misunderstanding with module_defaults units. We still need them to be part of output of this method. 
The issue was with order of execution of association and unassociation of module_defaults units which should be documented.

ubipop/__init__.py
```diff
@@ -132,6 +133,25 @@ def _filter_ubi_conf_list(self, config_list, content_sets, repo_ids):
 
         return filtered_conf_list
 
+    def expected_pulp_actions(self, content_sets=None, repo_ids=None):
```
rbikar: As this is intended as public API, can you at least add docstring here?

ubipop/__init__.py
```diff
@@ -497,6 +517,28 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
...
+
+        Association = namedtuple('Association', ['unit_id', 'type', 'src_repo', 'dest_repo'])
+        Unassociation = namedtuple('Unassociation', ['unit_id', 'type', 'dest_repo'])
+
+        associations, unassociations, _, _ = self._get_pulp_actions(*self._get_current_content())
+
+        for item in associations:
+            if item.units:
+                for unit in item.units:
+                    actions.associations.append(Association(unit.name, item.TYPE, item.src_repo.repo_id,
```
rbikar: unit.name here is not unique identifier of unit that is required for association call to pulp.
e.g. for rpm it's filename, for module it's nsvca. Probably instantiation of Association object with whole unit object would be ok.

ubipop/__init__.py
```diff
@@ -497,6 +517,28 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
...
+                                                            item.dst_repo.repo_id))
+
+        for item in unassociations:
+            if item.units:
+                for unit in item.units:
+                    actions.unassociations.append(Unassociation(unit.name, item.TYPE,
```
rbikar: see comment above, the same problem is here.

ubipop/__init__.py
```diff
@@ -497,6 +517,28 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
 
         return diff
 
+    def expected_pulp_actions(self):
```
rbikar: The name of this method is the same as in UbiPopulate class. It may be misleading.
Can you consider renaming it?

rbikar:
> @rbikar & everyone else, I won't be adding docs for this, as I feel it would require the creation of docs for all of UbiPopulate. So, I think this can be included in the "make public" issue that I've yet to create.

OK, but please add docstring for public method in UbiPopulate.

nathanegillett:
> Quick question here, why did you abandon #128 PR? Not a big deal but for other reviewers comments there can be valuable.

I agree it wasn't ideal. IIRC, there were problems reverting the work to make the requested changes and syncing this with git/hub.

nathanegillett:
> OK, but please add docstring for public method in UbiPopulate.

Will do! Thanks for the in-line feedback. Very helpful.

nathanegillett:
run tests

ubipop/__init__.py
```diff
@@ -497,10 +539,31 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
...
 
-    def run_ubi_population(self):
-        current_modules_ft, current_module_defaults_ft, current_rpms_ft, \
-            current_srpms_ft, current_debug_rpms_ft = self._get_current_content()
+    def export_pulp_actions(self):
+        """Formats and returns key data from expected pulp actions for easy external digestion"""
+
+        actions = ExpectedPulpActions([], [])
+
+        Association = namedtuple('Association', ['unit', 'type', 'src_repo', 'dest_repo'])
```
midnightercz: why did you add new class for Association action when there's already _utils.py AssociateAction?
And if so, why did you create those model classes nested?

nathanegillett: These un/association tuples are just used to format data from those objects. I suppose I can look into changing those classes to support this use. @rbikar, any objection to that?

nathanegillett: We can skip these namedtuples, I just think it may be more confusing for the caller of the API. Not too much of a difference, though.

nathanegillett:
@rbikar let me know what you think when you have a chance

nathanegillett:
run tests

nathanegillett:
run tests

nathanegillett:
run tests

nathanegillett:
Sorry, @release-engineering/content-delivery, I've performed a bunch of messy Git operations trying to troubleshoot broken CI which effectively erased the history of this PR. Problem is persisting.

nathanegillett:
Reported broken CI in #134

rbikar:
> @rbikar let me know what you think when you have a chance

@nathanegillett   @midnightercz 

Well I liked that simpler output that was there before.

One thing is that the PulpAction class and its subclasses won't exist in future after _pulp_client is replaced with pulplib, so we will need to do more refactoring here. So I guess creation of namedtuples or something similar should be better.

Second related thing is model classes in _pulp_client won't exists as well, so I guess namedtuple approach or similar will require less refactoring in future (or none if written in generic way) - also this lib will be coupled with Pub pushes so imho less changes -> less number of possible bugs.

midnightercz:
> Well I liked that simpler output that was there before.

You mean like string output? Then you can extend classes with to_str or something similar to get you format that you need. 

> Second related thing is model classes in _pulp_client won't exists as well, so I guess namedtuple approach or similar will require less refactoring in future
Right, but this PR is meant to provide public api for "getting expected actions on pulp". How would you then document that without having publicly documented actions as well?  

_UbiPopulate.expected_pulp_actions return array of some action. Internal structure of actions is unknown._

Isn't better to create simple public class with desired public api for example in models/*.py and once it's in pubtools_pulplib simply remove the model there and only import ModelClass from pubtools_pulplib?
Note that best would be to wait for implementation in pubtools_pulplib

rbikar:
> > Well I liked that simpler output that was there before.
> 
> You mean like string output? Then you can extend classes with to_str or something similar to get you format that you need.

Yeah, that would be even better - some export method in the PulpAction classes.


> 
> > Second related thing is model classes in _pulp_client won't exists as well, so I guess namedtuple approach or similar will require less refactoring in future
> > Right, but this PR is meant to provide public api for "getting expected actions on pulp". How would you then document that without having publicly documented actions as well?
> 
> _UbiPopulate.expected_pulp_actions return array of some action. Internal structure of actions is unknown._
> 
> Isn't better to create simple public class with desired public api for example in models/*.py and once it's in pubtools_pulplib simply remove the model there and only import ModelClass from pubtools_pulplib?
> Note that best would be to wait for implementation in pubtools_pulplib

Personally I would also wait for implementation in pubtools_pulplib so we don't have to do any extra refactoring work here or in Pub.

nathanegillett:
Okay, @rbikar, @midnightercz, we'll hold off on this until pubtools.pulplib is integrated.

PR #132, opened 2019-09-19
Revert "Conditional creation of install_requires of rpm distribution …

rbikar:
…(#114)"

Reverting this commit as it doesn't help with RPM build of this
project. This change is followed by another change in specfile
that manually removes rpm-py-installer from requirements.txt 
per recommendation in https://github.com/junaruga/rpm-py-installer/blob/master/docs/users_guide.md  (Q6)

See related DELIVERY-7245.

This reverts commit c8731e11e3de0dd53fef432b2c801e6bacab7315.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 requests
 more-executors
 ubi-config
+rpm-py-installer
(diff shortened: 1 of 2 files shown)


rbikar:
run tests

Issue #138, opened 2019-10-03
Use correct version of ubi config file for ubi repository population


rbikar:
#### Summary

In order to support new branching scheme for ubi config repos, we need to reflect that also in ubipop. In short, ubipop should select correct ubi config file based on version for given repo.
There are two scenarios for different types of repos:
**For DOT repositories:**
- select ubi config by platform_full_version repo note

**For mainline (e.g. 7Server) repositories**
- select ubi config by ubi_config_version note
- if repo note is not available, use default version ("7" for ubi7, "8" for ubi8 ) 

I both scenarios, if a config is not available for given version, fallback to default version.
Version of ubi config will available via version attr - https://github.com/release-engineering/ubi-config/issues/49

#### Issue Type
- Feature Request

#### Current behavior

Currently we don't support selecting ubi config file per version.

#### Desired behavior

- for every ubi repository correct ubi config is select by its version
- for determining which version should be used take platform_full_version repo note for DOT repositories and ubi_config_version repo note for mainline repositories 
- if ubi config is not available for given version, fallback to default version ("7" for ubi7, "8" for ubi8 ) 

#### Use case

Ubi DOT repositories are considered as testing repos, so we may want to used different (new) config file for for DOT repositories than for mainline repositories which are customer facing.

PR #140, opened 2019-10-15
Use correct version of ubi config file for population

JayZ12138:
This resloves #138
A new attr is added to Repo class and it will be populated while
searching repos.

After all config files are loaded, it now creates a config map, which
helps find the right config file for the repo to populate.


JayZ12138:
Tests will be added/modified later.

ubipop/__init__.py
```diff
@@ -132,18 +133,65 @@ def _filter_ubi_conf_list(self, config_list, content_sets, repo_ids):
...
                 _LOG.warning("Skipping current content triplet, some repos are missing")
                 continue
 
             for repo_set in repo_pairs:
-                UbiPopulateRunner(self.pulp, repo_set, config, self.dry_run,
+                # get the right config file by ubi_config_version attr, if it's None,
+                # then it's not a mainline repo, use platform_full_version instead.
+                version = repo_set.out_repos.rpm.ubi_config_version \
+                    or repo_set.out_repos.rpm.platform_full_version
+                right_config = self.ubiconfig_map.get(version, {}).get(config.filename)
```
rbikar: This doesn't seem to fulfill following acceptance criteria:
> if ubi config is not available for given version, fallback to default version ("7" for ubi7, "8" for ubi8 )

In ubi config repo there will always be 'default' branches ubi7 and ubi8 where you need to fallback if config  for required version isn't available. And if I read it correctly, it can assign None to right_config var which will lead ubipop to crash. This can happen for both DOT and mainline repos. 

Or there can be some "best effort" mechanism which can select the latest available config < required version, if required config is not available. But this is not needed now I guess.

JayZ12138: Thanks for this. I wasn't sure what's 'default version'.
I'm still confused about why 'version' could be None? could you elaborate on it?

JayZ12138: I mean, isn't platform_full_version always available?

JayZ12138: ah, I misread.
yeah, right_config could be None and it leads to the push crashing. I saw that in the first place and I didn't do any exception for it, because I think it should fail in this case, means there's a major error in config repo, we should fail this push. 
Do you think we should ignore and continue push?

JayZ12138:
I'm not sure why CI passed... 
I'm going to introduce fake loader to ubiconfig, which would make tests much easier.

JayZ12138:
run tests

ubipop/__init__.py
```diff
@@ -26,6 +26,10 @@ class RepoMissing(Exception):
     pass
 
 
+class ConfigMissing(Exception):
```
alexandrevicenzi: MissingConfig sounds better

ubipop/__init__.py
```diff
@@ -132,18 +137,84 @@ def _filter_ubi_conf_list(self, config_list, content_sets, repo_ids):
...
             for repo_set in repo_pairs:
-                UbiPopulateRunner(self.pulp, repo_set, config, self.dry_run,
+                ubi_config_version = repo_set.out_repos.rpm.ubi_config_version
+                platform_full_version = repo_set.out_repos.rpm.platform_full_version
+                # get the right config file by ubi_config_version attr, if it's None,
+                # then it's not a mainline repo, use platform_full_version instead.
+                # config file could also be missing for specific version, then the
+                # default config file will be used.
+                version = ubi_config_version or platform_full_version
+                right_config = self.ubiconfig_map\
```
alexandrevicenzi: Break only on `or`, this is hard to read, but not a big deal.

rbikar:
@JayZ12138 
Do you plan any changes in this PR after https://github.com/release-engineering/ubi-config/pull/57  is merged?

JayZ12138:
> @JayZ12138
> Do you plan any changes in this PR after [release-engineering/ubi-config#57](https://github.com/release-engineering/ubi-config/pull/57) is merged?

Yes.

JayZ12138:
run tests

JayZ12138:
run tests

JayZ12138:
Run tests

JayZ12138:
run tests

ubipop/__init__.py
nathanegillett: I think this is fine, conforms to existing scheme

PR #146, opened 2020-06-21
ci: fix pylint failures

rohanpm:
There was an attempt to disable a few warnings by comments, but that
wasn't working. It seems that pylint's behavior regarding multi-line
statements might have changed since the code was introduced.

Moving the comment prior to the statement works, so do that.


rbikar:
ok test

PR #147, opened 2021-05-26
Fix association of unwanted modular rpms [RHELDST-6453]

rbikar:
This fixes a situation when modular packages are selected for output set
of ubi repo but they shouldn't be there because respective modulemd unit
is not selected as a content of ubi repo.

This lead to situations when a repo contained modular rpm from a
modulemd that wasn't associated with repo. This is wrong and could lead
to problems with packages installation from ubi repos.

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -1348,3 +1348,23 @@ def test_get_pkgs_from_all_modules(mock_ubipop_runner):
     assert len(pkgs) == 2
     assert "tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm" in pkgs
     assert "tomcatjss-8.4.7-2.el8+1944+b6c8e16f.noarch.rpm" in pkgs
+
+
+def test_filter_pkgs_from_modules(mock_ubipop_runner):
+    mock_ubipop_runner.repos.modules['test5.30'] = [
+        get_test_mod(name="test", stream="5.30", packages=["tomcatjss-0:7.3.6-1.el8+1944+b6c8e16f.noarch"]),
+    ]
+
+    mock_ubipop_runner.repos.pkgs_from_modules['test5.30'] = [
+        get_test_pkg(
+            name="tomcatjss",
+            filename="tomcatjss-7.3.6-1.el8+1944+00000000.noarch.rpm",
+        ),
+        get_test_pkg(
+            name="tomcatjss",
+            filename="tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm",
+        )]
+
+    mock_ubipop_runner._filter_pkgs_from_modules()
+    assert len(mock_ubipop_runner.repos.pkgs_from_modules['test5.30']) == 1
+    assert mock_ubipop_runner.repos.pkgs_from_modules['test5.30'][0].filename == "tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm"
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -605,7 +605,7 @@ class UbiPopulateRunner(object):
         return diff
 
     def _filter_pkgs_from_modules(self):
-        regex = r'^\d+:'
+        regex = r'\d+:'
         reg = re.compile(regex)
 
         for name_stream, packages in self.repos.pkgs_from_modules.items():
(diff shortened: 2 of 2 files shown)


rbikar:
This waits for https://github.com/release-engineering/ubi-population-tool/pull/148

ubipop/__init__.py
```diff
@@ -604,6 +604,18 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
 
         return diff
 
+    def _filter_pkgs_from_modules(self):
+        regex = r'\d+:'
```
rohanpm, line 608: I'm a bit confused at the intent here.  I think it's to strip off the leading epoch from the RPMs, right?  Like:

  0:foo-1.2.3-4.el8.x86_64.rpm => foo-1.2.3-4.el8.x86_64.rpm

If so, does it need to be anchored at the beginning, making it:

```suggestion
        regex = r'^\d+:'
```

Without that, seems it could potentially match anywhere in the name which is brittle.

tests/test_ubipop.py
```diff
@@ -1348,3 +1348,23 @@ def test_get_pkgs_from_all_modules(mock_ubipop_runner):
...
+        get_test_pkg(
+            name="tomcatjss",
+            filename="tomcatjss-7.3.6-1.el8+1944+00000000.noarch.rpm",
+        ),
+        get_test_pkg(
+            name="tomcatjss",
+            filename="tomcatjss-7.3.6-1.el8+1944+b6c8e16f.noarch.rpm",
+        )]
+
+    mock_ubipop_runner._filter_pkgs_from_modules()
```
rohanpm, line 1368: This is OK, but it'd be nice if there was at least one test which tested this at a higher level too in order to show the overall impact.  The test here may protect against bugs in `_filter_pkgs_from_modules`, but it doesn't help ensure that the call to the method is wired up correctly.  e.g. a refactor or incorrect resolution of a git conflict which accidentally removes the call to `self._filter_pkgs_from_modules()`, would not be caught by any test if I understand right.

rohanpm (approved):

ubipop/__init__.py
rbikar: Yes, you're right, it's about stripping off the epoch.

tests/test_ubipop.py
rbikar: Yeah, no having high level test also lowers the code coverage.

rohanpm (approved):

ubipop/__init__.py
```diff
@@ -604,6 +604,18 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
 
         return diff
 
+    def _filter_pkgs_from_modules(self):
+        regex = r'^\d+:'
```
rbikar, line 608: @rohanpm 
Actually the epoch is not the leading thing there, but it's name of the package. 
E.g. tomcatjss-0:7.3.6-1.el8+1944+b6c8e16f.noarch 
So the anchor breaks the expected behavior now :(

ubipop/__init__.py
```diff
@@ -604,6 +604,18 @@ def _diff_lists_by_attr(self, list_1, list_2, attr):
...
 
+    def _filter_pkgs_from_modules(self):
+        regex = r'\d+:'
+        reg = re.compile(regex)
+
+        for name_stream, packages in self.repos.pkgs_from_modules.items():
+            # name_stream, packages == str, list of _pulp_client.Package
+
+            wanted_packages = list(chain.from_iterable([m.packages for m in self.repos.modules[name_stream]]))
+            wanted_packages = [reg.sub('', rpm_nevra) + '.rpm' for rpm_nevra in wanted_packages]
```
rohanpm, line 615: One other minor suggestion - this reg.sub would be a little safer if you gave it a count of 1 (do max 1 substitution).  Since we now remembered that the regex shouldn't be anchored at the beginning.

rohanpm (approved):

ubipop/__init__.py
rohanpm: Oops, my mistake.  Adding one other minor suggestion about count=1 in re.sub then (because anchor would have ensured it only matched once, now it's not guaranteed).

ubipop/__init__.py
rbikar: Yep, but I'll do it in another PR - the same substitution happens also in another place in code, I guess it would be better to fix both and ideally do it at the time when self.repos.modules is populated.
I'll merge this PR, we need to test it and deploy it ASAP, it looks it creates a lot of customer problems.
Travis seems to fail only on pylint issues, I'll fix those also in another PR.

PR #149, opened 2021-06-04
Improve CI

rbikar:
- removed py35,36,37 test env as it's already not supported or not needed
- added py38, py39 test envs
- added config for pylint


.travis.yml
```diff
@@ -9,11 +9,19 @@ matrix:
   include:
   - python: '2.7'
     env: TOX_ENV=py27
-  - python: '3.5'
-    env: TOX_ENV=static
   - python: '3.6'
```
rohanpm, line 12: If you *want* to keep the py3 versions older than 3.8 you certainly could, but from now I'm suggesting py3.8 as the oldest python3 version to be supported by default.  For pubtools projects it was recorded at https://release-engineering.github.io/pubtools/devguide.html#target-platforms .

The reasoning is that rhel8/ubi8 already provides python 3.8, so for any project migrating from py2 to py3 there doesn't seem to be any reason to use anything older than that.

rohanpm (approved):

rbikar:
I have just removed py36 and py37 envs, carrying over +1.

PR #152, opened 2021-06-29
Make matching modules and modular rpms efficient [RHELDST-6391]

rbikar:
This commit starts decoupling of big blob of code inside UbiPopulateRunner
class that did almost everything.

Methods for getting modulemd units and modular rpms and srpms was moved away
from this class to *Matcher class. It provides convenient methods for
searching in pulp and also encapsulates functionality of creation of
output sets of units for ubi repositories.

It was written in a way that is completely asynchronous and can be
easily extended with methods for matching e.g. non-modular units.

Currently ubipop is in hybrid state where two pulp clients are used
one from ubipop and the other from pubtools.pulplib. This makes
especially writing tests quite inconvenient.

### DIFF
--- a/tests/test_matcher.py
+++ b/tests/test_matcher.py
@@ -1,0 +1,608 @@
+import pytest
+
+from operator import attrgetter
+from more_executors.futures import f_proxy, f_return
+from pubtools.pulplib import (
+    RpmUnit,
+    Criteria,
+    YumRepository,
+    FakeController,
+    ModulemdUnit,
+)
+from ubiconfig import UbiConfig
+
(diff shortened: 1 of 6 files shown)


rohanpm (approved):
Some more DEBUG logs might be a good idea, I would imagine if this doesn't work as expected it might be a bit of a nightmare to debug right now.

rbikar:
> Some more DEBUG logs might be a good idea, I would imagine if this doesn't work as expected it might be a bit of a nightmare to debug right now.

Yeah, thanks for suggestion. I'll add some debug logs there.

PR #153, opened 2021-06-30
Optimize requests for repository

rbikar:
Previously ubipop requested repository from pulp
in every iteration in _create_srpms_output_set method.
Let's save the repository into dict in order not to do
unnecesary requests to pulp.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -482,7 +482,7 @@ class UbiPopulateRunner(object):
 
     def _create_srpms_output_set(self):
         rpms = chain.from_iterable(self.repos.packages.values())
-
+        binary_source_repo_map = {}
         for package in rpms:
             if package.sourcerpm is None:
                 _LOG.warning(
@@ -495,7 +495,9 @@ class UbiPopulateRunner(object):
                 if r.id == package.associate_source_repo_id
             ][0]
 
-            associate_src_repo = in_repo.get_source_repository()
+            associate_src_repo = binary_source_repo_map.setdefault(
+                in_repo.id, in_repo.get_source_repository()
+            )
 
             self.repos.source_rpms[package.name].append(
                 Package(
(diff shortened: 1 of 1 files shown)


nathanegillett (approved):

ubipop/__init__.py
```diff
@@ -495,7 +495,9 @@ def _create_srpms_output_set(self):
                 if r.id == package.associate_source_repo_id
             ][0]
 
-            associate_src_repo = in_repo.get_source_repository()
+            associate_src_repo = binary_source_repo_map.setdefault(
+                in_repo.id, in_repo.get_source_repository()
```
rohanpm, line 499: Hmm... I'm not entirely sure this works as you intended.

If we have this:

```
somedict.setdefault(somekey, expr)
```

The expression `expr` is evaluated every time even if the value ends up being unused.  So that means we're still calling get_source_repository the same number of times we were before.

If in_repo.get_source_repository() returns future then this probably still speeds up the code, since the function isn't blocking; but it would be doing pointless requests in the background, creating futures which are never used.  That might have some negative performance impact too because those requests do make use of the limited number of request threads for querying Pulp, so it might delay later operations on the client.

We probably just need to use a boring old 'if' statement here.

rohanpm (changes requested):

ubipop/__init__.py
rbikar: hmm, ok. I thought that setdefault immediately returns value, if the key exists.

rohanpm: Well it's just a plain function.  If you imagine you implemented a function:

```python
def my_fn(a, b):
    # ...

# called like this:
my_fn(1+2, 3+4)
```

Is there any code you can write inside of my_fn which can reach "upwards" and cause (3+4) not to be evaluated?

No, it's too late by that point, python had to calculate a value for your function's arguments before your function could be called.

rohanpm (approved):

PR #154, opened 2021-07-21
Fix futures consistency in output repositories [RHELDST-6391]

rbikar:
One of output repository object was stored as YumRepository which wasn't
desirable because other code expects this to be Future.

Let's make this consistent now have all repositories stores as futures.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -9,7 +9,7 @@ from pubtools.pulplib import Client, Criteria, PublishOptions
 import ubiconfig
 
 from more_executors import Executors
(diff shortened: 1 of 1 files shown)


rbikar:
run tests

rbikar:
run tests

rbikar:
It looks that integration tests are properly triggered but there are two issues:
1) Problem with matching profiles in modulemd - will be fixed by https://github.com/release-engineering/ubi-population-tool/pull/155
2) Missing repo note in testing data - filed RHELDST-7249

rohanpm (approved):

PR #155, opened 2021-07-21
Fix KeyError when profile is missing in modulemd

rbikar:
When ubi-config requests a profile from modulemd but
the profile is not in the modulemd, ubipop previously failed
with KeyError.

Now missing profile is not treated as a fatal error.

### DIFF
--- a/tests/test_matcher.py
+++ b/tests/test_matcher.py
@@ -328,6 +328,42 @@ def test_modular_rpms_filenames_per_profiles(ubi_config):
     )
 
 
+def test_modular_rpms_filenames_per_profiles_missing_profile(ubi_config):
+    """Test getting filename from module artifacts, request for non-existing profile in modulemd"""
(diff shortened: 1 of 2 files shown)


rbikar:
run tests

rohanpm (approved):
Any comments about the failing Jenkins CI?  Anyway, I'm approving despite that since I don't have a good idea whether that CI is generally working.

rbikar:
> Any comments about the failing Jenkins CI? Anyway, I'm approving despite that since I don't have a good idea whether that CI is generally working.

There is a known issues that will be fixed by https://github.com/release-engineering/ubi-population-tool/pull/154
But there is also problem with testing data that will be fixed within  RHELDST-7249

This CI hasn't run properly for almost two year, but given the fact that logic in ubipop is still the same, it should work after testing data are fixed.

PR #164, opened 2021-09-07
Fix creation of source RPMs output set

rbikar:
Previously the method _get_rpm_output_set() stripped away
all modular packages and kept only the latest version of packages.

This seems to be too strict for source RPMs in some edge cases, at the
time when source RPMs output set is created - because the output set is
based on final output sets of binary and debug rpms that are already properly
filtered - they include only the latest versions of RPMs and non-modular
packages. Because of that we don't need to check again if the source RPM is
modular or not, or keep the latest version of SRPM. Only blacklisting
should happen. There are some edge cases when the original implementation
stripped some source RPMs that should be part of the output set. It's
probable that the edge cases happen due to data inconsistency especially
for modular packages.

We always need to include source RPMs of all binary and debug RPMs
unless they're blacklisted.

### DIFF
--- a/tests/test_matcher.py
+++ b/tests/test_matcher.py
@@ -825,6 +825,82 @@ def test_get_rpm_output_set(ubi_config):
     assert output[0].version == "11"
 
 
+def test_get_rpm_output_set_include_modular_and_all_versions(ubi_config):
+    """tests getting rpm output set from RpmMatcher, modular packages included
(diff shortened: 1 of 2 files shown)


rbikar:
run tests

rbikar:
run tests

rbikar:
Jenkins failure relates to RHELDST-7382.

nathanegillett (approved):

PR #167, opened 2021-09-10
Rework matching modulemd_defaults units [RHELDST-6393]

rbikar:
Matching modulemd_defaults units for ubi repos population is now
reworked and uses Client from pubtools-pulplib in the similar way
as matching other units (modulemd, rpm ...) works.


rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

PR #170, opened 2021-09-15
Refactored getting current content from ubi repos

rbikar:
This comming includes refactoring of getting current content
of ubi repositories.
- Queries for getting content from ubi repos now re-uses code from
  Matcher class. Some private methods were made public and changed to
  class methods.
- All related queries located originally located in _pulp_client.py
  were removed also with model classes. It brings us closer to
  100% usage of pubtools-pulplib Client. Currently we still need to use
  only calls from _pulp_client.py for content un/association that are
  not ready in pubtools-pulplib at the time of creation of this commit.
- Added couple of helper methods for test and updated/add tests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ ubi-config>=2.2.0
 rpm-py-installer
 pyrsistent<0.17; python_version < '3'
 pubtools-pulplib>=2.9.0
+attrs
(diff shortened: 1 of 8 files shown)


rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

PR #176, opened 2022-01-10
Update tox cov job to generate xml coverage report

rajulkumar:
Codecov action to upload coverage report was upgraded to v2.
However, the reports were not uploaded correctly as v2 accepts
coverage reports in xml format only. Hence, updating the cov
job in tox to generate coverage reports in xml instead of html.


rbikar:
run tests

rbikar (approved):

PR #179, opened 2022-03-30
Integrate ubi-manifest [RHELDST-8706]

rbikar:
This change introduces uasge of ubi-manifest that provides
manifests of final content of ubi repos. Following
is included in this changed:
* implemented a simple client for ubi-manifest servivce with task poller
* integrated usage of ubi-manifests in the ubipop workflow
* added required cli option --ubi-manifest-url
* tests were updated

There is still old unused code - it will be cleaned in subsequent
PRs.

### DIFF
--- a/ubipop/ubi_manifest_client/client.py
+++ b/ubipop/ubi_manifest_client/client.py
@@ -1,0 +1,125 @@
+import logging
+import os
+import threading
+from concurrent.futures import as_completed
+
+import requests
+from more_executors import Executors, f_map, f_proxy
+
+from .models import UbiManifest
+
+LOG = logging.getLogger(__name__)
+
+API = "api/v1"
+
+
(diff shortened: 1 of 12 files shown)


rbikar:
Note: Jenkins CI is NOT expected to work at this stage as it requires update to integration tests and environment.

ubipop/ubi_manifest_client/client.py
```diff
@@ -0,0 +1,125 @@
...
+            response.raise_for_status()
+            raise
+
+        response.raise_for_status()
+        return out
+
+    @property
+    def _session(self):
+        if not hasattr(self._tls, "session"):
+            self._tls.session = requests.Session()
```
rohanpm, line 60: I suppose this might be coming later, but I'll point it out now.  It'll probably need some authentication mechanism.

rohanpm (approved):

PR #182, opened 2022-05-13
Fix handling modular units

rbikar:
- added missing nsvca property for ModulemdUnit model class
- added extra search for modulemd_defaults units
-- for association or unassociation actions they are diffed by profiles
   so we need to get this info from pulp. Name or stream is not enough.


rbikar:
run tests

rbikar:
run tests

rohanpm (approved):

PR #183, opened 2022-05-18
Fixed ModulemdUnit model class

rbikar:
Attribute version on ModulemdUnit class has to be stored as integer,
otherwise any queries on pulp will not work as expected.

### DIFF
--- a/ubipop/ubi_manifest_client/models.py
+++ b/ubipop/ubi_manifest_client/models.py
@@ -82,7 +82,7 @@ class ModulemdUnit(UbiCompatibleUnit):
         return cls(
             name=n,
             stream=s,
-            version=v,
+            version=int(v),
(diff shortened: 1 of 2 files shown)


rajulkumar (approved):

ubipop/ubi_manifest_client/models.py
```diff
@@ -82,7 +82,7 @@ def from_nsvca(cls, nsvca, src_repo_id, dst_repo_id):
         return cls(
             name=n,
             stream=s,
-            version=v,
+            version=int(v),
```
rohanpm, line 85: Note if you did this via an attrs 'converter' instead, then it would reliably happen also for instances created without using from_nsvca.

rohanpm (approved):

PR #185, opened 2022-06-02
Refactor ubi integration test

LinglingXiang:
1. run pop for ubi7
2. add rpm related test cases
3. run pop for ubi8


tests/integration/test_association.py
```diff
@@ -402,289 +334,334 @@ def assert_pkgs(pkgs_included, pkgs_found):
...
-                md.profiles for md in cfg.modules.whitelist if md.name == md_name
-            ][0]
+    return Klass
 
-            for profile in profiles:
-                pkgs_profile = module["profiles"].get(profile, [])
 
-                for pkg in pkgs_profile:
-                    assert pkg in pkgs_found, '"{}" not in the repository.'.format(pkg)
+def get_in_repo_from_config(config_file):
```
rbikar, line 471: This can be de-duplicated as it contains almost the same code as in get_out_repo_from_config()

tests/integration/test_association.py
```diff
@@ -402,289 +334,334 @@ def assert_pkgs(pkgs_included, pkgs_found):
...
+                    continue
+
+            filenames.add(filename)
+
+    return filenames
+
+
+@pytest.mark.skipif(INTEGRATION_NOT_SETUP, reason="Integration test is not set up.")
+def test_run_ubi_pop_for_rhel8():
+    run_ubipop_tool("ubi8/rhel-8-server.yaml")
```
rbikar, line 638: Hmm, what exactly does this test case checks?

LinglingXiang: actually this test case is needed for the next module test cases, so put it here

tests/integration/test_association.py
```diff
@@ -2,34 +2,52 @@
...
 PULP_USER = os.getenv("TEST_PULP_USER")
 PULP_PWD = os.getenv("TEST_PULP_PWD")
 PULP_CERT_PATH = os.getenv("TEST_PULP_CERT_PATH")
 PULP_SECURE = os.getenv("TEST_PULP_SECURE", "0").lower() in ["true", "yes", "1"]
+MANIFEST_URL = os.getenv("TEST_MANIFEST_URL")
+REQUESTS_CA_BUNDLE = os.getenv("REQUESTS_CA_BUNDLE")
 PULP_RPM_REPO_PREFIX = "/pulp/rpmrepos/"
 
-TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "./data/ubi7")
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "./data")
```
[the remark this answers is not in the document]
LinglingXiang: [~] hi Robert, yes, i used the same source for ubipop and ubi-manifest. the data/ubi7/xxxx is copied from the gitlab source which used in ubi-manifest url, also the ubi8. here copied it, it just used for get the expect result.eg: some rpm, some module and some module default

LinglingXiang:
hi Robert, 
_We need explicitly defined expected results of tests (even before it wasn't ideal with those tons of asserts in code). I would like to have some json or yaml files with expected results (rpm, modulemd, modulemd_defaults content of the populated repos) that will be asserted as a whole after ubipop run. Also an input data are a bit mysterious to me (I only know that the pulp DB is restored with some db dump)._

_You shouldn't use code directly from ubipop (with some exceptions) or use copied code for sorting or filtering. Definitely not from Matcher classes which are going away in near future._

yeap, let's go back to the test plan. 
My thought is:
expected:
query from the input repo and get some rpms, 
Acutal:'
query from the output repo to check has the same rpm or not
use the dnf/yum from the pulp if have

your idea is :
expected:
just define  an whole expected result file
actual:
then query from the output repo(ubi), and match if it have
use the dnf/yum from the pulp if have



_It looks that the most problematic thing here is config handling because we need to use only the GitLab source, I need to think about this more. I believe it can handled in some way that doesn't need drastic changes in the integration tests._
for this, seem you misunderstand something. the config data in ubi7 folder and ubi8 is the same as gitlab source that used in the ubi-manifest. why i copied the same under here, it use for getting the expected result and verifying.

tests/integration/test_association.py
LinglingXiang: yeap, we can merge the two method to one

tests/integration/test_association.py
```diff
@@ -251,106 +273,25 @@ def separate_modules(module_list):
...
-    debug_repos = get_repos_from_cs(cfg.content_sets.debuginfo.output)
-
-    repos = chain(rpm_repos, srpm_repos, debug_repos)
-
-    for repo in repos:
-        assert_empty_repo(repo)
-
-
-@pytest.mark.skipif(INTEGRATION_NOT_SETUP, reason="Integration test is not set up.")
-def test_modulemd_defaults_unit_copy():
```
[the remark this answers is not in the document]
LinglingXiang: yeap, i agree some of test cases didn't have yum or dnf checking, i'll add it.

tests/integration/test_association.py
```diff
@@ -2,34 +2,52 @@
...
 
-TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "./data/ubi7")
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "./data")
 
 INTEGRATION_NOT_SETUP = PULP_HOSTNAME is None
 
 
 def load_ubiconfig(filename):
     loader = ubiconfig.get_loader(TEST_DATA_DIR)
+    print (loader.load(filename))
```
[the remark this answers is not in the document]
LinglingXiang: ack

tests/integration/test_association.py
rbikar: Tests should be independent of each other. If it's meant to do some test setup, than it should named and called before or inside the test case.

LinglingXiang:
draft this comment

PR #191, opened 2022-08-18
Add cache purging [RHELDST-12390]

rbikar:
Added cache purge for mutable urls of ubi repositories
using fastpurge client.

### DIFF
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -6,6 +6,7 @@ import os
 import shutil
 import sys
 import tempfile
+import textwrap
 
 import pytest
 import ubiconfig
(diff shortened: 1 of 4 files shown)


rbikar:
Integration tests seem not to be in a workable shape still.

tests/test_ubipop.py
```diff
@@ -1221,9 +1224,32 @@ def test_get_current_content(mock_ubipop_runner, pulp, skip_debug_repo):
     assert source_rpms[0].name == "test-srpm"
 
 
+@pytest.fixture(name="fastpurge_env")
+def fixture_fastpurge_env():
+    with tempfile.NamedTemporaryFile("w") as fp:
+        EDGERC_CFG = fp.name
```
rohanpm, line 1230: Minor: it stands out as odd that these are named with all caps.  My first thought was, the intent here is to overwrite some global variables during the test, but it can't be as there's no 'global' statement.  So they are just local variables and it's really weird to name those with all caps.

It might be connected with a later comment about hacking the attributes on ubi_populate instance.  I guess this fixture in theory could just be using monkeypatch.setenv("UBIPOP_EDGERC_CFG", "<the value>") etc, and then the code would be working a bit more consistent with how it works when the tool is run for real.

tests/test_ubipop.py
```diff
@@ -1289,6 +1315,10 @@ def test_populate_ubi_repos(get_debug_repository, get_source_repository, request
         ubi_manifest_url="https://ubi-manifest.com",
     )
 
+    edgerc_path, fastpurge_root_url = fastpurge_env
+    ubi_populate._EDGERC_CFG = edgerc_path
```
rohanpm, line 1319: You don't actually need to hack these attributes, do you?  Because you made the code read from environment variables, so you could also just set those environment variables before you create the object which might be cleaner since that's the way the code is expected to work "for real".

monkeypatch fixture can take care of temporarily setting the environment variables during the test.

rohanpm (dismissed):

rohanpm (approved):

PR #193, opened 2022-10-14
Enable logging of ubi-config [RHELDST-6398]

rbikar:
For CLI this change enables logging of ubi-config on INFO level.


rbikar:
run tests

rohanpm (approved):

PR #194, opened 2022-10-31
Adding logs and exception for ubi repo sets [RHELDST-11863]

amcmahon-rh:
If the debug or source repo for an RPM repo was not present, the returned value was None, which caused an attribute exception since the code would try accessing values from None. This change checks to ensure these repos actually exist, and raises an exception if they're not present.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -86,7 +86,8 @@ class UbiRepoSet(object):
             _LOG.error("Input Source repo does not exist")
 
         if not self.in_repos.debug:
-            _LOG.warning("Input Debug repo does not exist")
+            fatal = True
+            _LOG.error("Input Debug repo does not exist")
 
         if not self.out_repos.rpm:
             fatal = True
@@ -97,7 +98,8 @@ class UbiRepoSet(object):
             _LOG.error("Output Source repo does not exist")
 
         if not self.out_repos.debug:
-            _LOG.warning("Output Debug repo does not exist")
+            fatal = True
+            _LOG.error("Output Debug repo does not exist")
 
         if fatal:
             raise RepoMissing()
@@ -419,7 +421,10 @@ class UbiPopulate(object):
         return ubi_repo_sets
 
     def _get_population_sources(self, out_repo):
-        if not out_repo.population_sources:
+        if (
+            not hasattr(out_repo, "population_sources")
+            or not out_repo.population_sources
+        ):
             raise PopulationSourceMissing
 
         repos = [
--- a/tests/test_ubipop.py
+++ b/tests/test_ubipop.py
@@ -289,6 +289,17 @@ def test_get_population_sources_empty():
         )  # pylint: disable=protected-access
 
 
+def test_get_population_sources_none():
+    fake_ubipopulate = FakeUbiPopulate(
+        "foo.pulp.com", ("foo", "foo"), False, ubiconfig_dir_or_url=TEST_DATA_DIR
+    )
+
+    with pytest.raises(PopulationSourceMissing):
+        fake_ubipopulate._get_population_sources(
+            None
+        )  # pylint: disable=protected-access
+
+
 @patch("pubtools.pulplib.YumRepository.get_debug_repository")
 @patch("pubtools.pulplib.YumRepository.get_source_repository")
 def test_get_ubi_repo_sets(get_debug_repository, get_source_repository):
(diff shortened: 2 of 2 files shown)


rbikar:
run tests

rbikar:
run tests

ubipop/__init__.py
```diff
@@ -399,7 +399,17 @@ def _get_ubi_repo_sets(self, ubi_binary_cs):
         ubi_repo_sets = []
         for out_rpm_repo in rpm_repos:
             out_source_repo = out_rpm_repo.get_source_repository()
+            if not out_source_repo.result():
```
rbikar, line 402: Please consider de-duplication of this like:
```
for repo, repo_type in zip([r1, r2], ["source", "debug"]):
...
```

ubipop/__init__.py
```diff
@@ -399,7 +399,17 @@ def _get_ubi_repo_sets(self, ubi_binary_cs):
...
+                _LOG.warning(
+                    "Source repo corresponding to {} not found".format(out_rpm_repo.id)
+                )
+                raise RepoMissing()
             out_debug_repo = out_rpm_repo.get_debug_repository()
+            if not out_debug_repo.result():
+                _LOG.warning(
+                    "Debug repo corresponding to {} not found".format(out_rpm_repo.id)
+                )
+                raise RepoMissing()
```
rbikar, line 412: Missing debug repo is not fatal error, it shouldn't raise exception.
Can also check `UbiRepoSet._ensure_repos_existence()` method? Seems like we check existence for this two times now. Maybe some re-ordering or refactoring of checks can be useful.

rbikar (changes requested):

ubipop/__init__.py
amcmahon-rh: 1. The _get_ubi_repo_sets() function is called in a try/except block. If the repo doesn't exist this exception is caught and the content set is skipped. In fact, when looking at the UbiRepoSet class, if the repo doesn't exist the exception is also raised, that is, if the execution were to carry on, the same exception would be raised already

2. Another approach to this would be to changing _get_population_sources to return None if the input is None and let UbiRepoSet be the one to raise an error. I think the check inside the UbiRepoSet should be left in place, in case the class is used by other routes at some point. Stuff like this makes me miss monads

rbikar: The main point of my comment is that missing **debug** repo is not a problem and in this case the processing can continue. At least this was specified when ubi tooling was designed but in practice this never happened.
On the other hand missing **source** repo is a problem and we can't continue.

rbikar: I checked briefly newer code and it looks that debug repo is required is some places already.
This is gonna fail, if the debug repo is missing, e.g. `out_repos.debug` is `Future[None]` or something that returns None.
```
        # start async querying for modulemds and modular and non-modular packages
        if self.ubim_client:
            binary_manifest = self.ubim_client.get_manifest(self.repos.out_repos.rpm.id)
            debug_manifest = self.ubim_client.get_manifest(
                self.repos.out_repos.debug.id
            )
```
So I think we could switch to logic where the debug repo is required.

rbikar: But I also think we don't want to extend the scope of the issue and improving the logging properly is enough right now.

amcmahon-rh: erh, I didn't see this final comment. I've implemented the approach I mentioned in the second point in my previous comment. It prevents the attribute access exception and a lets the UbiRepoSet class take care of checking for existence. I've also changed the debug repo not existing to be "fatal". LMK if this approach makes more sense

ubipop/__init__.py
```diff
@@ -419,6 +421,9 @@ def _get_ubi_repo_sets(self, ubi_binary_cs):
         return ubi_repo_sets
 
     def _get_population_sources(self, out_repo):
+        if not out_repo:
```
rbikar, line 424: Unfortunately it looks this part of code is poorly covered by tests (which is not your fault) and in real use it likely won't work as intended because out_repo will be future in cases: `out_source_repo` and `out_debug_repo` . 
I guess `if out_repo.result():` should fix the issue but `out_rpm_repo` has to be made a future sooner that it is now.

Just checked the `get_source_repository()` and  `get_debug_repository()` methods in pubtools-pulplib and they will return either a proxy future returning repo or future returning None, but even then the future is evaluated as True in if statement:
```
>>> from more_executors.futures import f_return
>>> bool(f_return())
True
>>> bool(f_return().result())
False
>>> 
```

rbikar (changes requested):

ubipop/__init__.py
amcmahon-rh: would it be possible instead to extend the original check with `hasattr`. So:

`if not hasattr(out_repo, "population_sources") or not out_repo.population_sources`

will produce false for None because the first condition fails, which means the second condition doesn't get evaluated.

rbikar: does `hasattr` work with proxy futures?

amcmahon-rh: It appears to work for proxy futures yes. I've tested it on Ling's QE environment

rbikar: ok, let's try `hasattr` but please make sure it's covered by unit tests.

rbikar:
codecov will be fixed by https://github.com/release-engineering/ubi-population-tool/pull/195

rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

rbikar:
run tests

rbikar (approved):

PR #196, opened 2022-11-09
Legacy code in ubipop is removed   [RHELDST-13337]

jihupdc:
1. legacy code in ubipop is removed.
2. ubi-manifest is the only source of expected content for ubipop.


rbikar (changes requested):
There is way more code for deletion.
Please check also _matcher.py, _pulp_client.py and _utils.py.
I believe most of code in _matcher.py can go away.

PR #197, opened 2022-11-10
Improve the log of associate and unassociate [RHELDST-9945]

LinglingXiang:
1.add method to make log for associate and unassociate more readable
2.add the unit test for testing the unassocate module default log
3.add the unit test for testing the associate package log

### DIFF
--- a/ubipop/_pulp_client.py
+++ b/ubipop/_pulp_client.py
@@ -137,6 +137,16 @@ class Pulp(object):
     def _rpms_query(self, rpms):
         return [{"filename": rpm.filename} for rpm in rpms]
 
+    def _filter_item_log(self, type_ids, unit):
+        log_message = ""
+        if "rpm" in type_ids or "srpm" in type_ids:
+            log_message = unit.filename
+        elif "modulemd" in type_ids:
+            log_message = unit.nsvca
+        elif "modulemd_defaults" in type_ids:
+            log_message = ":".join((unit.name, unit.stream))
+        return log_message
+
     def unassociate_units(self, repo, units, type_ids):
         url = "repositories/{dst_repo}/actions/unassociate/".format(dst_repo=repo.id)
         data = {
@@ -145,10 +155,10 @@ class Pulp(object):
                 "filters": {"unit": {"$or": self._get_query_list(type_ids, units)}},
             },
         }
-        log_msg = "Unassociating %s from %s"
+        log_msg = "Unassociating %s(%s) from %s"
         for unit in units:
-            _LOG.info(log_msg, str(unit), repo.id)
-
+            info = self._filter_item_log(type_ids, unit)
+            _LOG.info(log_msg, ",".join(type_ids), info, repo.id)
         ret = self.do_request("post", url, data).json()
         return [task["task_id"] for task in ret["spawned_tasks"]]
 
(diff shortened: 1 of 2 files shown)


ubipop/_pulp_client.py
```diff
@@ -165,9 +180,15 @@ def associate_units(self, src_repo, dest_repo, units, type_ids):
                 },
             },
         }
-        log_msg = "Associating %s from %s to %s"
+        log_msg = "Associating %s(%s) from %s to %s"
         for unit in units:
-            _LOG.info(log_msg, str(unit), src_repo.id, dest_repo.id)
+            if "rpm" in type_ids or "srpm" in type_ids:
```
rbikar, line 185: I just wonder, could you use the same approach for "Unassociate" log? The method `_filter_item_log()` looks redundant and too dependent on implementation detail of `UbiUnit class`. Please note that there is also `UbiCompatibleUnit` used for Unassociate action and it doesn't provide `_unit` attr (which shouldn't be accessed because it's private).

ubipop/_pulp_client.py
```diff
@@ -165,9 +180,15 @@ def associate_units(self, src_repo, dest_repo, units, type_ids):
...
+                info = unit.filename
+            if "modulemd" in type_ids:
+                info = unit.nsvca
+            if "modulemd_defaults" in type_ids:
+                info = ":".join((unit.name, unit.stream))
+            _LOG.info(log_msg, type(unit).__name__, info, src_repo.id, dest_repo.id)
```
rbikar, line 191: I believe that you can just use `type_ids` or something like `",".join(type_ids)` instead of accessing __name__ : `type(unit).__name__`

ubipop/_pulp_client.py
```diff
@@ -137,6 +142,16 @@ def _module_defaults_query(self, module_defaults):
     def _rpms_query(self, rpms):
         return [{"filename": rpm.filename} for rpm in rpms]
 
+    def _filter_item_log(self, unit):
+        log_message = ""
+        if type(unit._unit) == ModulemdUnit:
```
rbikar, line 147: 1. You shouldn't access private attr in object.
2. More standard way of checking type is `instance()`
3. This looks like a redundant method, please see other comments.

ubipop/_pulp_client.py
```diff
@@ -165,9 +180,15 @@ def associate_units(self, src_repo, dest_repo, units, type_ids):
                 },
             },
         }
-        log_msg = "Associating %s from %s to %s"
+        log_msg = "Associating %s(%s) from %s to %s"
         for unit in units:
-            _LOG.info(log_msg, str(unit), src_repo.id, dest_repo.id)
+            if "rpm" in type_ids or "srpm" in type_ids:
+                info = unit.filename
+            if "modulemd" in type_ids:
```
rbikar, line 187: It would be better for perf. if you use `if-elif-else` with some reasonable default for `info` var.

rbikar (changes requested):

rbikar:
1. Please rewrite PR and commit description so it fits into the limit.
2. Also a bit more precise commit message wouldn't harm.
3. Can you also add some simple test or update the current ones? Thx.

rbikar:
@LinglingXiang 
Changes look OK, please rebase on current master to fix the py27 tests. Thx.

rbikar (approved):

PR #198, opened 2022-11-16
Fixup: still dumps exception instead of simple log

amcmahon-rh:
The main issue was resolved, in that we were trying to access a value which wasn't present. However, we should return None and allow the UbiRepoSet class to raise a RepoMissing exception,  which is already captured and logged.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -425,7 +425,7 @@ class UbiPopulate(object):
             not hasattr(out_repo, "population_sources")
             or not out_repo.population_sources
         ):
-            raise PopulationSourceMissing
+            return []
 
         repos = [
             self.pulp_client.get_repository(repo_id)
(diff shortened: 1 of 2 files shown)


amcmahon-rh:
An alternative approach would be to keep the exception and add PopulationSourceMissing to the caught exceptions. However, allowing the UbiRepoSet class to take care of it will give logs on which repo specifically is missing.

ubipop/__init__.py
```diff
@@ -425,7 +425,7 @@ def _get_population_sources(self, out_repo):
             not hasattr(out_repo, "population_sources")
             or not out_repo.population_sources
         ):
-            raise PopulationSourceMissing
+            return None
```
rbikar, line 428: Can you make it return empty list instead so we retutrn the same types in all cases?

rbikar (changes requested):

amcmahon-rh:
run tests

rbikar:
@amcmahon-rh 
Please rebase on current master, there is py27 CI fix.

rbikar (approved):

PR #202, opened 2023-01-20
Fix CI

rbikar:
Removed tox config for travis which caused issues + we stopped using travis long time ago.


rbikar:
run tests

rbikar:
not needed anymore

PR #207, opened 2023-05-11
Fix tox.ini

rbikar:
On older tox versions, passenv doesn't work with passenv variables specified inline with separator (,).


rbikar:
run tests

rohanpm (approved):

PR #208, opened 2023-05-22
Retire python 2 support [RHELDST-15344]

jihupdc:
1.Remove python2-specific entries in 'classifiers'. 2.Specify minimal python version to 3.8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,14 +32,12 @@ setup(
         "Development Status :: 5 - Production/Stable",
         "Intended Audience :: Developers",
         "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.6",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -32,14 +32,12 @@ def get_requirements():
...
         "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.6",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
     url="https://github.com/release-engineering/ubi-population-tool",
     install_requires=get_requirements(),
+    python_requires='>=3.8',
```
rbikar, line 40: I don't think this should change to `3.8`, ubipop jobs run currently on `python3.7`

rbikar (changes requested):

PR #210, opened 2023-05-24
Fix CDN and Fastpurge client handling

rbikar:
Previously CDN or Fastpurge client were instatiated even if there was not enough configuration provided for them.

Also the edgerc conf. file was required everytime even if we didn't want to use cache purging functionality.

Let's now fix that so we get clients only if at least some minimal configuration is provided.


rbikar:
run tests

rohanpm (approved):

PR #214, opened 2023-06-02
Fix handling ARL templates

rbikar:
Previously even if an empty string was used as ARL templates, it was enough to enable querying CDN headers. Let's now fix that, we query CDN only we have valid ARL templates.


rbikar:
run tests

rohanpm (approved):

PR #215, opened 2023-06-07
Fix futures chaining

rbikar:
Previous usage of `f_map` caused creation of nested future and subsequent `result()` call didn't really wait for nested futures to finish. This caused immediate end of whole program before all submitted futures were started.

In production env likely only some of submitted cdn cache purges were allowed to start, some paths were not purged at all.

Let's now use `f_flat_map` instead which can be used with future-returning functions and any nesting of futures is automatically avoided - therefore the program will wait for all submitted cache purges.

### DIFF
--- a/ubipop/_cdn.py
+++ b/ubipop/_cdn.py
@@ -10,7 +10,7 @@ from concurrent.futures import as_completed
 import requests
 from fastpurge import FastPurgeClient
 from more_executors import Executors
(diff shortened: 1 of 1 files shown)


rbikar:
run tests

rajulkumar (approved):

rohanpm (approved):

PR #217, opened 2023-10-12
Fix TTL extraction [RHELDST-20510]

drepelov:
Previously the TTL extraction was implemented in such way that it could potentially extract the wrong value, if the URL path contained a component that also matched the regex.
This commit changes the re.match() to re.search(), which ensures it will find the first occurrence of the match.
It also changes whitelist_externals to allowlist_externals in tox.ini


rbikar:
run test

jm-wk (approved):

rbikar:
run test

rbikar:
ok to test

rbikar:
ok to test

rbikar (approved):

rbikar:
The jenkins CI doesn't seem work at the moment.
Merging anyway.

PR #218, opened 2023-10-26
Sanitize provided ARL templates

rbikar:
ARL templates are extracted from environment variable which can in practice easily include some trailing whitespace characters. Unstripped ARLs with trailing whitechars are not accepted by the FastPurge API resulting in 4xx errors.


rbikar:
ok to test

rohanpm (approved):

rajulkumar (approved):

PR #219, opened 2023-10-29
Fix setting verify for requests

rbikar:
Due to bug https://github.com/psf/requests/issues/3829, setting 'verify' on session doesn't work if REQUESTS_CA_BUNDLE is set on environment. The bundle defined via REQUESTS_CA_BUNDLE will take precedence and the custom bundle provided via 'verify' is ignored.

Let's now set 'verify' for each request.

This can be revetrted when the bug is fixed, likely in python-requests-v3.

### DIFF
--- a/ubipop/_cdn.py
+++ b/ubipop/_cdn.py
@@ -216,6 +216,10 @@ class CdnClient:
         return self._tls.session
 
     def _head(self, *args, **kwargs):
+        # set verify for each request
+        # verify set on session doesn't work due to https://github.com/psf/requests/issues/3829
+        # if REQUESTS_CA_BUNDLE is set on env, it takes precedence
+        kwargs["verify"] = self._session.verify
         return self._session.head(*args, **kwargs)
 
     def _on_failure(self, header, exception):


ubipop/_cdn.py
```diff
@@ -216,6 +216,10 @@ def _session(self):
         return self._tls.session
 
     def _head(self, *args, **kwargs):
+        # set verify for each request
+        # verify set on session doesn't work due to https://github.com/psf/requests/issues/3829
+        # if REQUESTS_CA_BUNDLE is set on env, it takes precedence
+        kwargs["verify"] = self._session.verify
```
rohanpm, line 222: I'm not certain which is better, but there is also a `trust_env` var you can set on the session to `False`.

Lately my approach has been: if the Session will have either `verify` or proxies used, then always set explicitly (verify, proxies, trust_env=False).

rohanpm (approved):

ubipop/_cdn.py
rbikar: you approach sounds also good, I guess I'll keep it simple for the time being.

PR #221, opened 2024-01-03
Make publishing of repos optional [RHELDST-21951]

drepelov:
After the rhel-ubi repos sync, this tool also took care of publishing the affected ubi repos to the CDN. However, this did not work for Exodus. It was decided that all publishes should be done via Pub, which is already configured and works for Exodus.

This commit prepares this tool for removal of publishing by making it possible to switch the publishing on/off as needed.

It also removes the intended deduplication of srpms in `_get_pulp_actions_src_pkgs()` which was a bit confusing and also unnecessary, since deduplication is done during manifest creation. Finally, it removes some forgotten unused variables and renames some variables so it's more understandable.

### DIFF
--- a/ubipop/__init__.py
+++ b/ubipop/__init__.py
@@ -115,7 +115,7 @@ class UbiPopulate(object):
         ubiconfig_dir_or_url=None,
         verify=True,
         workers_count=4,
-        output_repos=None,
+        output_repos_file=None,
         **kwargs
     ):
         # legacy client implemeted in this repo, it's expected to be replaced by pubtools.pulplib.Client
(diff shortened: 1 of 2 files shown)


ubipop/__init__.py
```diff
@@ -329,37 +312,34 @@ def populate_ubi_repos(self):
...
             ubi_binary_repos.extend(
-                [repo_set.out_repos.rpm.id for repo_set in repo_pairs]
+                [repo_set.out_repos.rpm.id for repo_set in repo_sets]
             )
 
         with UbimClient(self._ubi_manifest_url) as ubim_client:
             tasks = ubim_client.generate_manifest(ubi_binary_repos)
             tasks.result()
 
-            with Publisher(**self._publisher_args) as publisher:
```
rbikar, line 347: Because this is quite a big change for ubipop, do you think you can add some env. var that would decide if we run ubipop with or without publish?  I am asking for this because we are going to deploy new version to all environments at once and we would like to ensure that the overall solution works on stage first and then switch it also on prod. In case of some problems we would be able to switch it back to original code with publish.

The clean-up commit can be part of another PR which we will merge after some soak time.

rbikar (changes requested):

rbikar (approved):

PR #223, opened 2024-03-14
Replace usage of _pulp_client with pubtools-pulplib [RHELDST-141]

negillett:
This commit removes the library's native _pulp_client ("legacy") and replaces it with usage of pubtools-pulplib methods for associating and unassociating content.

Additionally, PulpActions were replaced with simple abstractions for grouping content to be associated/unassociated.

Tests were removed/edited as necessary.


negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
I'm at a total loss as to why we're seeing error 111 on '/api/v1/manifest'. I've touched nothing in the ubi client.

rbikar:
> I'm at a total loss as to why we're seeing error 111 on '/api/v1/manifest'. I've touched nothing in the ubi client. Any ideas, @rbikar?

I think the ubi-manifest service in the test env was in a bad shape, I re-ran tests and they passed OK.

ubipop/__init__.py
```diff
@@ -537,46 +532,74 @@ def run_ubi_population(self):
...
+                criteria = self._criteria_for_units(u.units, u.unit_type)
+                unassociation_fts.append(dst_repo_obj.remove_content(criteria))
+        return unassociation_fts
+
+    def _criteria_for_units(self, units, unit_type):
+        unit_crit = None
+        partial_crit = []
+        for unit in units:
+            if unit_type == "rpm":
+                unit_crit = Criteria.with_unit_type(RpmUnit)
```
rbikar, line 577: `unit_crit` is loop invariant, you can set this outside of the loop. It may need a little redesign of this method. You can e.g. pass the class (e.g. `RpmUnit`)  via argument.

ubipop/__init__.py
```diff
@@ -537,46 +532,74 @@ def run_ubi_population(self):
...
+        for u in unassociations:
+            # It's reasonable that the destination repo may not contain the
+            # units targeted for un/association so let's plan for that.
+            if u.units:
+                dst_repo_obj = self.pulp_client.get_repository(u.dst_repo.id)
+                criteria = self._criteria_for_units(u.units, u.unit_type)
+                unassociation_fts.append(dst_repo_obj.remove_content(criteria))
+        return unassociation_fts
+
+    def _criteria_for_units(self, units, unit_type):
```
rbikar, line 572: Do you think you can some batching mechanism for criteria? During bigger events like GA, criteria tend to be pretty big (especially for rpms) so it would be nice to have associations or unassociations split into batches.

ubipop/__init__.py
```diff
@@ -537,46 +532,74 @@ def run_ubi_population(self):
...
+                    "arch": unit.arch,
+                }
+                for m_part, value in nsvca_dict.items():
+                    m_crit.append(Criteria.with_field(m_part, value))
+                partial_crit.append(Criteria.and_(*m_crit))
+            if unit_type == "modulemd_defaults":
+                unit_crit = Criteria.with_unit_type(ModulemdDefaultsUnit)
+                partial_crit.append(Criteria.with_field("name", unit.name))
+
+        assert unit_crit
```
rbikar, line 596: I don't think those asserts are needed, this code is not public API, so the code + tests should ensure that it works properly.

Btw, currently it won't say much to the user if it fails on this.

ubipop/__init__.py
```diff
@@ -537,46 +532,74 @@ def run_ubi_population(self):
...
+        assert partial_crit
+        return Criteria.and_(unit_crit, Criteria.or_(*partial_crit))
+
+    def _wait(self, futures):
         for ft in as_completed(futures):
-            tasks = ft.result()
-            if tasks:
-                self.pulp_client.wait_for_tasks(tasks)
+            ft.result()
```
rbikar, line 602: This probably can a bit re-written.
Inside the the loop, the future `ft` is already completed because we loop using `as_completed`.

I'd suggest either to use just `wait(futures)` ,  or maybe also add some logging that would tell us what copy/removal of content was completed, e.g. via `future.add_done_callback(some_logging_function)`

rbikar (changes requested):

ubipop/__init__.py
negillett: Yeah, accepting the unit object was my initial approach but I thought it'd make the function easier to use if it accepted a string instead. You're right that this is needlessly reassigning though so I'll improve that.

ubipop/__init__.py
negillett: This was probably leftover from debugging. Thanks for pointing it out!

ubipop/__init__.py
negillett: Ah, okay. I think we'll just loop fts for `result` then. Using `wait` would force us to support the `not_done` cases, and there doesn't seem to be much benefit there. Adding a logging callback would be nice but I don't think it would make sense here, maybe upon future creation.

negillett: Decided against implementing a log callback as there's no useful info available at the future level such as the function name or arguments supplied. We can log future results in other ways if we really want to do that.

ubipop/__init__.py
negillett: Assuming you mean something like returning a list of criteria rather than a single criteria for all the given units, e.g., a set of rpm criteria, modulemd criteria, and modulemd_defaults criteria. Yeah, that should be rather easy.

tests/test_ubipop.py
```diff
@@ -1312,22 +1229,27 @@ def test_populate_ubi_repos(
...
+        for unit in repo_content:
+            if isinstance(unit, ModulemdUnit):
+                assert unit.nsvca == new_modulemd.nsvca
+            if isinstance(unit, ModulemdDefaultsUnit):
+                assert unit.profiles == new_modulemd_defaults.profiles
 
-    assert repo_ids_published == expected_published_repo_ids
-    # unfortunately we can't check actual content od repos because
-    # un/associate calls are using custom client not the pubtools-pulplib Client
-    # TODO add check for actual content after we move to pubtools-pulplib Client
```
negillett, line 1253: Added this bit to check the content on the published repos but the results seem odd to me. The rpm isn't published  and debug and source repos are empty. Is that intended here, @rbikar?

ubipop/__init__.py
rbikar: yeah, no problem. It's not in the scope of the issue.

ubipop/__init__.py
rbikar: It can be problematic especially with RPMs, returning criteria e.g. for 100 units and running copy/remove and then another 100 units would be useful.

ubipop/__init__.py
```diff
@@ -383,27 +364,111 @@ def __init__(
...
+            self.log_pulp_actions(
+                associations + (mdd_association,),
+                unassociations + (mdd_unassociation,),
+            )
+        else:
+            fts = []
+            fts.extend(self._do_associations(associations))
+            fts.extend(self._do_unassociations(unassociations))
+            for ft in fts:
+                ft.result
```
rbikar, line 403: this doesn't look OK.
You need to call the result method:
```
ft.result()
```

negillett: Good catch. Ty!

tests/test_ubipop.py
negillett: Thanks for confirming, I'll try to track down that rpm unit (still missing after url mocks removed).

ubipop/__init__.py
negillett: Oh, I see. I've split the criteria by unit but I can work batch size limit into the un/associations methods.

tests/test_ubipop.py
negillett: Update:
Added some logs to see what was going on, looks like pulp_client.copy_content isn't handling RpmUnits properly.

> Got an RPM unit: RpmUnit(src_repo_id='input_binary', dst_repo_id='ubi_binary', filename='golang-2.a.x86_64.rpm')
Made criteria for RPM: ((content_type_id IN ['rpm', 'srpm']) AND filename=='golang-2.a.x86_64.rpm')
Did associations: [Task(id='e3e70682-c209-4cac-629f-6fbed82c07cd', completed=True, succeeded=True, tags=[], repo_id='input_binary', units=[ModulemdUnit(name='test_md', stream='s', version=200, context='c', arch='x86_64')], units_data=[]), Task(id='82e2e662-f728-b4fa-4248-5e3a0a5d2f34', completed=True, succeeded=True, tags=[], repo_id='input_binary', units=[], units_data=[])]

Continuing investigation

negillett: Or maybe there's some reason filename isn't findable on fake_pulp.

rbikar: I found the culprit, the fake_pulp.copy_content() by default requires signing_key set on Rpm units.
So e.g. setting `signing_key="xxxx"` on testing RPM unit will fix the problem.

negillett: Oh, nice. Thanks, @rbikar!

negillett:
run tests

tests/test_ubipop.py
negillett: @rbikar, I submitted https://github.com/release-engineering/pubtools-pulplib/pull/219 too. No reason to behave differently here.

negillett:
run tests

negillett:
run tests

tests/test_ubipop.py
```diff
@@ -1312,22 +1216,30 @@ def test_populate_ubi_repos(
...
-        assert len(publish.tasks) == 1
-        assert publish.tasks[0].completed
-        assert publish.tasks[0].succeeded
-
-    assert repo_ids_published == expected_published_repo_ids
-    # unfortunately we can't check actual content od repos because
-    # un/associate calls are using custom client not the pubtools-pulplib Client
-    # TODO add check for actual content after we move to pubtools-pulplib Client
+        repo = publish.repository
+        repo_content = repo.search_content().result().data
```
rbikar, line 1223: This looks a bit suspicious,
I believe you can do just `repo_content = repo.search_content()` and iterate over it.
Or you can get the content with convenient methods:
https://release-engineering.github.io/pubtools-pulplib/api/pulpcore.html#pubtools.pulplib.Repository.rpm_content and others

ubipop/__init__.py
```diff
@@ -626,55 +637,73 @@ def log_pulp_actions(self, associations, unassociations):
...
+                dst_repo_obj = self.pulp_client.get_repository(u.dst_repo.id)
+                for chunk in list(self._batch_units(u.units)):
+                    criteria = self._criteria_for_units(chunk, u.unit_type)
+                    unassociation_fts.append(dst_repo_obj.remove_content(criteria))
+        results = []
+        for ft in unassociation_fts:
+            results.extend(ft.result())
+        return results
+
+    def _batch_units(self, units):
```
rbikar, line 674: Can you please add test that actually will use more than one batch?

ubipop/__init__.py
```diff
@@ -626,55 +637,73 @@ def log_pulp_actions(self, associations, unassociations):
...
+        for i in range(0, len(units), self._action_batch_size):
+            yield units[i : i + self._action_batch_size]
+
+    def _criteria_for_units(self, units, unit_type):
+        partial_crit = []
+        for unit in units:
+            if unit_type is RpmUnit:
```
rbikar, line 681: can you change this multiple ifs section into if-elif style?

tests/test_ubipop.py
```diff
@@ -1246,6 +1163,7 @@ def test_populate_ubi_repos(
         arch="x86_64",
         filename="golang-1.a.x86_64.rpm",
         sourcerpm="golang-1.a.x86_64.src.rpm",
+        signing_key="xxxx",
```
rbikar, line 1166: minor: the signing_key can hold a bit better value like `test_signing_key`

tests/test_ubipop.py
```diff
@@ -1312,22 +1216,30 @@ def test_populate_ubi_repos(
     ubi_populate.populate_ubi_repos()
     history = fake_pulp.publish_history
 
-    # there should be 3 repositories succesfully published
+    # there should be 3 repositories successfully published
     assert len(history) == 3
-    expected_published_repo_ids = set(["ubi_binary", "ubi_debug", "ubi_source"])
```
rbikar, line 1317: In production we already stopped publishing content via this tool, but isn't this part of test still valid?

rbikar (changes requested):

tests/test_ubipop.py
negillett: I haven't seen an existing variable yet and there doesn't seem to be much benefit to creating one since the value isn't being used. I can make the key look a little more realistic though I suppose.

ubipop/__init__.py
negillett: Yeah, thanks. Not sure what started this habit but there's no use for the extra if checks here.

ubipop/__init__.py
```diff
@@ -626,55 +637,73 @@ def log_pulp_actions(self, associations, unassociations):
...
-        if self.repo_set.out_repos.debug.result():
-            current_debug_rpms = f_proxy(
-                self._executor.submit(
-                    Matcher.search_rpms, criteria, [self.repo_set.out_repos.debug]
+    def _do_associations(self, associations):
+        association_fts = []
+        for a in associations:
+            for src_repo, units in a.src_repo_to_unit_map.items():
+                if units:
+                    src_obj = self.pulp_client.get_repository(a.get_repo(src_repo).id)
```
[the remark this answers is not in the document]
negillett: Yeah, that's right. Thanks

ubipop/_utils.py
```diff
@@ -1,108 +1,31 @@
...
-        for unit in self.units:
-            src_repo_unit_map.setdefault(unit.associate_source_repo_id, []).append(unit)
-
-        return src_repo_unit_map
-
-    def _get_repo_obj(self, repo_id):
+        self._src_repo_to_unit_map = None
+
+    @property
+    def src_repo_to_unit_map(self):
```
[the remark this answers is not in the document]
negillett: Done!

ubipop/__init__.py
```diff
@@ -626,55 +637,73 @@ def log_pulp_actions(self, associations, unassociations):
...
-            self._executor.submit(
-                Matcher.search_srpms, criteria, [self.repo_set.out_repos.source]
-            )
-        )
-
-        if self.repo_set.out_repos.debug.result():
-            current_debug_rpms = f_proxy(
-                self._executor.submit(
-                    Matcher.search_rpms, criteria, [self.repo_set.out_repos.debug]
+    def _do_associations(self, associations):
```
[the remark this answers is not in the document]
negillett: Done!

ubipop/__init__.py
```diff
@@ -626,55 +637,73 @@ def log_pulp_actions(self, associations, unassociations):
...
+                if units:
+                    src_obj = self.pulp_client.get_repository(a.get_repo(src_repo).id)
+                    dst_obj = self.pulp_client.get_repository(a.dst_repo.id)
+                    for chunk in list(self._batch_units(units)):
+                        criteria = self._criteria_for_units(chunk, a.unit_type)
+                        association_fts.append(
+                            self.pulp_client.copy_content(src_obj, dst_obj, criteria)
+                        )
+        results = []
+        for ft in association_fts:
```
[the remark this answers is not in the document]
negillett: Done!
Wasn't sure how crucial it was for rpm+modulemd units to be done together.

tests/test_ubipop.py
negillett: Yeah, seems the result call isn't necessary anymore (it was at one point, unsure why now).
I was hesitant about using the *_content repo attributes since they each call search_content, but it does look a lot better in the test.

ubipop/__init__.py
negillett: Creating another large test with 100s of unit objects sounds expensive for a unit test. Something of that scale should be captured in the integration tests already, right?

However, I did move this method to utilities and added a unit test for it.

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
@rbikar, is ubi-manifest okay? I keep seeing connection errors (Connection aborted) for ubi-manifest task in `test_ubipop_not_filter_module_rpm_with_different_version`.

However, I think the test is failing because it's expected "common [d]" in the module profile but the profiles are "rhel" and "common". Should that "[d]" really be expected?

tests/integration/test_association.py
```diff
@@ -646,7 +630,7 @@ def test_ubipop_not_filter_module_rpm_with_different_version(pulp_client):
         mod_name == "container-tools"
     ), "Expected modulemd: httpd, found modulemd: {}".format(mod_name)
     assert (
-        "common [d]" in mod_profile
+        "common" in mod_profile
```
negillett, line 633: "rhel" and "common" are present, neither has this "[d]" mark.
@rbikar, I'm assuming it's meant to label the default profile but I don't see a mechanism for that in this repo which may have changed and wonder if that's still used/expected.

negillett:
run tests

tests/integration/test_association.py
rbikar: Yeah, the `[d]` should mean that the profile is the default one.
It's possible that something changed either in the testing modulemd/modulemd_defaults content or in yum/dnf tooling.

And for ubi-manifest health - I don't manage/check this testing instance, you would to check with our QEs.

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
run tests

negillett:
@rbikar, @rohanpm, this is ready for review.
I'm still unsure why some modulemd profiles no longer contain "[d]" to signal default but I don't believe the change was introduced here (it was only ever present in the integration tests). It could be a difference between rhel7 and rhel8 content--this is a key difference between the two integration tests that asserted the presence of the mark.

rbikar:
> @rbikar, @rohanpm, this is ready for review. I'm still unsure why some modulemd profiles no longer contain "[d]" to signal default but I don't believe the change was introduced here (it was only ever present in the integration tests). It could be a difference between rhel7 and rhel8 content--this is a key difference between the two integration tests that asserted the presence of the mark.

In realistic usage the ubi7 shouldn't have any modules (but I guess in test the situation can be different).

Anyway you need to examine `modulemd_defaults` units in the test instance of rhsm-pulp before and after the test.

I checked current real ubi8 image and e.g. for perl module I get this:
Output of `dnf module list` (I was unable to copy this output as it shows on my terminal)
```
perl                 5.24       common [d Practical Extraction and Report Language                                                                                                                                                             
                                ], minima 
                                l         
perl                 5.26 [d]   common [d Practical Extraction and Report Language                                                                                                                                                             
                                ], minima 
                                l         
perl                 5.30       common [d Practical Extraction and Report Language                                                                                                                                                             
                                ], minima 
                                l         
perl                 5.32       common [d Practical Extraction and Report Language                                                                                                                                                             
                                ], minima 
                                l   
``` 
and corresponding `modulemd_defaults` unit in production rhsm-pulp data:
(Maybe the testing code reads wrongly the output as in terminal the formatting doesn't look very nice)
```
...
            "name": "perl",
            "profiles": {
                "5.24": [
                    "common"
                ],
                "5.26": [
                    "common"
                ],
                "5.30": [
                    "common"
                ],
                "5.32": [
                    "common"
                ]
            },
            "pulp_user_metadata": {},
            "repo_id": "ubi-8-for-x86_64-appstream-rpms__8",
            "stream": "5.26"
        },
...
```
which say 2 things:
* what profile is default for which stream
* what stream is default for the module

so it should be dependent on the data in pulp only, can you double-check the removal and copy of modulemd_defaults units? There might some subtle bug (or the formatting problem as I mentioned above).

negillett:
run tests

negillett:
"common [d]" is showing up now for that one tests 🤷🏻

negillett:
run tests

negillett:
@rohanpm @rbikar ptal

PR #232, opened 2024-06-20
Retire obsolete Akamai cache purge support

rajulkumar:
Exodus-gw is now responsible for Akamai CDN cache flush during publish in all the supported workflows. It makes the code for CDN cache flush obsolete. Hence, remove the code pertaining to CDN cache flush.


rajulkumar:
coverage job fails with "Error: Codecov token not found. Please provide Codecov token with -t flag." which doesn't makes sense as the job uses codecov@v4

PR #233, opened 2024-06-21
Retire obsolete Akamai cache purge support

rajulkumar:
Exodus-gw is now responsible for Akamai CDN cache flush during publish in all the supported workflows. It makes the code for CDN cache flush obsolete. Hence, remove the code pertaining to CDN cache flush.

### DIFF
--- a/ubipop/_cdn.py
+++ b/ubipop/_cdn.py
@@ -1,90 +1,30 @@
-# inspired by CDN client and cache purge implementation
-# in https://github.com/release-engineering/pubtools-pulp/
 import logging
-import os
-import re
-import threading
-from collections import namedtuple
-from concurrent.futures import as_completed
 
-import requests
-from fastpurge import FastPurgeClient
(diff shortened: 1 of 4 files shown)


rajulkumar:
@negillett @drepelov Can you please review this patch? Unable to invite as reviewers here.

negillett:
run tests

negillett (approved):

ubipop/_cdn.py
```diff
@@ -97,211 +37,3 @@ def _enqueue(repo):
     def wait_publish_and_purge_cache(self):
```
drepelov, line 37: Just a nitpick: Since this method no longer actually does the cache purge, wouldn't it be clearer to just call it `wait_publish`?

rajulkumar: Done

drepelov (approved):

rbikar (approved):

PR #234, opened 2024-07-04
Authenticate session with kerberos

drepelov:
To access ubi-manifest server, requests session now needs to be authenticated using kerberos.
The `requests-kerberos` library used for the authentication needs some base packages to be installed in order to work properly. Therefore this commit adds an `Install base packages` step in `tox-test.yml` otherwise the tests would fail.

Note: on Jenikins the tool runs on `fedora-jenkins-minimal` image, where those packages are already installed so I think no changes there should be necessary.


drepelov:
run tests

rbikar (changes requested):

rbikar:
run tests

rbikar:
@drepelov 

As agreed, taking overt this PR.
I've added a couple of things that should good enough for further deployment with kerb. auth.

I don't expect jenkins tests to pass.

PR #237, opened 2024-08-30
Add the command to let pub client to do the publish [RHELDST-25071]

LinglingXiang:
1.let pub client to do the publish

### DIFF
--- a/tests/integration/test_association.py
+++ b/tests/integration/test_association.py
@@ -365,9 +365,31 @@ def filter_debug_packages(type, cfg_packages_whitelist):
     return pkgs_expected_exclude
 
 
+def run_pub_client_publish(repo_ids, target):
+    cmd = ["pub", "publish", "--target", target]
+    for repo_id in repo_ids:
+        cmd.extend(["--repo", repo_id])
+    try:
+        ret = subprocess.check_output(cmd, stderr=subprocess.STDOUT).decode("utf-8")
+        if "CLOSED" not in ret:
+            logging.info("The publish is not successful, so please check it first")
+            raise Exception("The publish fails as some reason")
+        else:
+            logging.info("The publish is successful")
+    except subprocess.CalledProcessError as ex:
+        logging.error(
+            "Command: %s returns with error %s: %s",
+            ex.cmd,
+            ex.returncode,
+            ex.output,
+        )
+        raise ex
+
+
 @pytest.mark.skipif(INTEGRATION_NOT_SETUP, reason="Integration test is not set up.")
 def test_run_ubi_pop_for_rhel7(pulp_client):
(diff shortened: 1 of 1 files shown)


LinglingXiang:
@drepelov @rbikar , please help review it, thanks

LinglingXiang:
@drepelov @rbikar , the format issue fixed,  please help review it, thanks

tests/integration/test_association.py
```diff
@@ -365,9 +365,30 @@ def filter_debug_packages(type, cfg_packages_whitelist):
...
 
+def run_pub_client_publish(repo_ids, target):
+    cmd = ["pub", "publish", "--target", target]
+    for repo_id in repo_ids:
+        cmd.extend(["--repo", repo_id])
+    try:
+        ret = subprocess.check_output(cmd, stderr=subprocess.STDOUT).decode("utf-8")
+        if "CLOSED" not in ret:
+            print("The publish is not successful, so please check it first")
+            raise Exception(ret)
```
rbikar, line 376: Not a big deal, it may be better to raise more specific exception.

tests/integration/test_association.py
```diff
@@ -365,9 +365,30 @@ def filter_debug_packages(type, cfg_packages_whitelist):
...
 
 
+def run_pub_client_publish(repo_ids, target):
+    cmd = ["pub", "publish", "--target", target]
+    for repo_id in repo_ids:
+        cmd.extend(["--repo", repo_id])
+    try:
+        ret = subprocess.check_output(cmd, stderr=subprocess.STDOUT).decode("utf-8")
+        if "CLOSED" not in ret:
+            print("The publish is not successful, so please check it first")
```
rbikar, line 375: seems like you sometimes use `print` and sometimes `logging` it would good to use only one variant, I guess.

tests/integration/test_association.py
```diff
@@ -365,9 +365,30 @@ def filter_debug_packages(type, cfg_packages_whitelist):
...
+            raise Exception(ret)
+        else:
+            print("The publish is successful")
+    except subprocess.CalledProcessError as ex:
```
rbikar, line 379: you probably want to re-raise this exception, otherwise tests will continue with unpublished repos.

rbikar (changes requested):

tests/integration/test_association.py
LinglingXiang: fixed

tests/integration/test_association.py
LinglingXiang: fixed

tests/integration/test_association.py
LinglingXiang: fixed

rbikar (approved):

PR #240, opened 2024-09-23
Fix the failture of ubipop for rhel7

LinglingXiang:
After the slave update from fedora 38 to fedora 40, the module query list return empty string, so remove the empty string and make the test case is pass.

### DIFF
--- a/tests/integration/test_association.py
+++ b/tests/integration/test_association.py
@@ -427,14 +427,18 @@ def test_run_ubi_pop_for_rhel7(pulp_client):
         ), "Can't find the modulemd_default :{}".format(modulemd_defaults)
 
     # verify the modulemd exist and modulemd default in the cdn server
-    modulemds = query_repo_modules(None, "ubi", rpm_repo_url, full_data=True)
+    modulemds = query_repo_modules(None, "ubi", rpm_repo_url)
+    while "" in modulemds:
+        modulemds.remove("")
     assert len(modulemds) == 3, "Unexpected repo modulemds found."
+    # assert the module prefile
+    modulemds = query_repo_modules(None, "ubi", rpm_repo_url, full_data=True)
     mod_name, mod_profile = separate_modules(modulemds)[0]
     assert mod_name == "httpd", "Expected modulemd: httpd, found modulemd: {}".format(
         mod_name
     )
     assert (
-        "common [d]" in mod_profile
+        "common" in mod_profile
     ), "Modulemd httpd should have common profile as default."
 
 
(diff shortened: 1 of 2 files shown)


LinglingXiang:
@rbikar @drepelov , could you help review it, thanks

tests/integration/test_association.py
```diff
@@ -427,14 +427,18 @@ def test_run_ubi_pop_for_rhel7(pulp_client):
         ), "Can't find the modulemd_default :{}".format(modulemd_defaults)
 
     # verify the modulemd exist and modulemd default in the cdn server
-    modulemds = query_repo_modules(None, "ubi", rpm_repo_url, full_data=True)
+    modulemds = query_repo_modules(None, "ubi", rpm_repo_url)
```
rbikar, line 430: you do another unnecessary call for the second time a few lines later, can you please refactor this so it does only one call and then filters received data in ` query_repo_modules` ?

tests/integration/test_association.py
```diff
@@ -427,14 +427,18 @@ def test_run_ubi_pop_for_rhel7(pulp_client):
...
+    while "" in modulemds:
```
rbikar, line 431: can you move this to `query_repo_modules` function? This filtering shouldn't be part of test. You can likely need this operation for other tests in future, so it's better to be done once in the function.

tests/integration/test_association.py
```diff
@@ -427,14 +427,18 @@ def test_run_ubi_pop_for_rhel7(pulp_client):
...
     assert len(modulemds) == 3, "Unexpected repo modulemds found."
+    # assert the module prefile
+    modulemds = query_repo_modules(None, "ubi", rpm_repo_url, full_data=True)
     mod_name, mod_profile = separate_modules(modulemds)[0]
     assert mod_name == "httpd", "Expected modulemd: httpd, found modulemd: {}".format(
         mod_name
     )
     assert (
-        "common [d]" in mod_profile
+        "common" in mod_profile
```
rbikar, line 441: Did really the output changed for default modules? Or do we have rather problem in testing data or ubi-population-tool?

rbikar (changes requested):

tests/integration/test_association.py
LinglingXiang: hi Robert, the cause is the same as previous one query_repo_modules. After some investigation,  the method works well in fedora38, in fedora 40, it will return in wrong format
Name      Stream                Profiles                   Summary             
httpd     2.4 [d]               common [ Apache HTTP Server  
                                d], defa                     
                                ult, dev                     
                                el   
if i update the yum update util-linux in fedora 40, the method can return mormally. 
In summary, nothing wrong in this method, but something problem in os.
so i will change the fedora 40 slave script instead of changing the method

tests/integration/test_association.py
LinglingXiang: the same as below comments, i will close the merge request.
and submit a patch for changing the slave

PR #242, opened 2024-11-21
Authenticate with requests-gssapi [RHELDST-25389]

Gdetrane:
Migrate deprecated requests_kerberos library to requests_gssapi for kerberos authentication.

PR #243, opened 2024-11-29
Retire native ubipop publishing [RHELDST-25072]

Gdetrane:
+ Remove `cdn.py` module containing the `Publisher` class
+ Remove the *UBIPOP_SKIP_PUBLISH* env variable used for switching between publishers
+ Removed the *publisher* option for classes using it


rbikar:
run tests

ubipop/__init__.py
```diff
@@ -259,18 +253,12 @@ def populate_ubi_repos(self):
         with UbimClient(self._ubi_manifest_url) as ubim_client:
             tasks = ubim_client.generate_manifest(ubi_binary_repos)
             tasks.result()
-
-            if self._skip_publish != "false":
-                _LOG.warning("Repository publishing via ubipop will be skipped")
-                self._run_ubi_population(
-                    repo_sets_list, out_repos, ubim_client, publisher=None
+            if os.environ.get("UBIPOP_SKIP_PUBLISH"):
```
rbikar, line 256: I don't think we need to warn. It's never been documented anywhere.

poetry.lock
rbikar, line 1: Although it's not a big deal, but it looks that changes in the lock file are not relevant to the PR.  At least I don't see any change in direct deps right now in PR.

rbikar (changes requested):
Almost OK, please rebase (it should likely fix the lock file)

ubipop/__init__.py
Gdetrane: I felt I was probably being too verbose with an unnecessary warning, my fear was that some people might try to set the variable thinking the old implementation is still there and see no effects, but I guess if someone knows about it in the first place, this should not be a concern.
I'll remove it right away

poetry.lock
Gdetrane: I was having issues with the tests and regenerating the lock file fixed them, I wasn't sure whether to commit it or not.
I rebased this fork's branch on upstream/master on the last push, that does not seem to resolve the conflict though, maybe I didn't do it properly with the incoming changed during rebase?
I'm still in doubt about the lock file itself and whether I should've committed in the first place, because technically without that change the test suite wouldn't even install for all envs if I remember correctly

rbikar: Unfortunately we haven't set up automated deps update (and update to lock file), but if you can, can you please create extra PR with update to all deps and then we can rebase this PR.

rbikar: Conflict there is likely because of https://github.com/release-engineering/ubi-population-tool/pull/242 - which had new deps.

Gdetrane: Oh right, we moved from requests-kerberos to requests-gssapi explicitly, but I'm surprised that the different pyproject.toml did not show up during rebase (I did it after #242 was merged)?

Anyways, would you mean simply having a fork from master explicitly with a new lock file generated by the updated pyproject.toml? That one should already be in master so it shouldn't be too much of an effort

rbikar:
@Gdetrane 
I am not sure why diff still shows changes in lock file and pyproject.toml...

Gdetrane:
@rbikar should be okay now

rbikar:
run tests

rbikar:
@Gdetrane 
yep, diff is OK now, but can you check the code coverage report? It seems like we get a bit lower number.

rbikar (approved):

PR #259, opened 2025-04-01
Update CI

rbikar:
* add py313 tests
* uses correct python version for py312 which was updated by renovate accindentally


rbikar:
run tests

drepelov (approved):

PR #263, opened 2025-04-08
Update deps and minimal python

rbikar:
* python39 is now minimal
* updated manually the lock file for poetry


rbikar:
run tests

PR #274, opened 2025-06-20
Fix usage of tox and poetry in ubi-population-tool [RHELDST-32220]

jihupdc:
Install poetry in system-wide with pip or pipx.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,9 +3,9 @@ envlist = py39, py312, py313, static, docs, integration
 
 [testenv]
 commands = pytest -v {posargs}
+allowlist_externals = sh, poetry
 skip_install = true
-commands_pre = 
-	pip install poetry
+commands_pre =
 	poetry run pip -V
 	poetry install --with dev,test
 
@@ -16,12 +16,11 @@ passenv =
 	TEST_MANIFEST_URL
 	GITLAB_CONFIG_URL
 	PUB_CONFIG_FILE
-allowlist_externals = yum
+allowlist_externals = yum, poetry
 commands =
 	pytest -v {posargs} tests/integration
 
 [testenv:static]
-allowlist_externals = sh
 commands =
 	black --check .
 	sh -c 'pylint -f colorized ubipop tests; test $(( $? & (1|2|4|32) )) = 0'
@@ -33,7 +32,6 @@ commands =
 
 [testenv:docs]
 use_develop = true
-allowlist_externals = sphinx-build
 commands  =
 	poetry install --with docs
 	sphinx-build -M html docs docs/_build


drepelov:
run tests

tox.ini
```diff
@@ -21,7 +21,7 @@ commands =
 	pytest -v {posargs} tests/integration
 
 [testenv:static]
-allowlist_externals = sh
+allowlist_externals = sh, poetry
```
Gdetrane, line 24: It wouldn't hurt to have this here, however I don't think it's necessary for this test env since poetry is not being invoked.

Everything else looks good 👍

Gdetrane: I'm not sure I explained myself with this

drepelov: Yes, you are right, the subenvs are inheriting the directives of the base testenv - so we do not need to specify it in each test subenv unless we need something additional there. 
Also it seems that `sphinx-build` in `docs`  is not needed to be explicitly allowed at all (at least locally) - that's what confused me.

So @jihupdc can you please just add the poetry to the `allowlist_externals` for `integration` test, and remove the `allowlist_externals` altogether from all the other tests?

drepelov: @jihupdc Looks like you didn't do any of that. You need to add poetry also to the `allowlist_externals` for the integration test (under `[testenv:integration]` ). And you need to remove the whole line with `allowlist_external = ` from `[testenv:static]` and `[testenv:docs]` . The `[testenv]` should stay as you have it now.

Gdetrane: @jihupdc I think the problem is that you should remove `allowlist_externals` altogether, whereas I still see `allowlist_externals = sh` in the static env. 
My understanding is that if you add `allowlist_externals` to any env, it won't **append** extra arguments to the original list, but it will overwrite it entirely.

So, when you do `allowlist_externals = sh` on the static env, you're removing poetry from it. Same thing with the integration test env, where you have `allowlist_externals = yum`, which in turn is going to remove the other dependencies like poetry and sh.

I think you should remove allowlist_externals from static and in integration, since you add a new dependency, you need to rewrite the entire list, e.g. `allowlist_externals = sh, poetry, yum`

rbikar:
looks this change broke integration tests:
```
integration: freeze> python -m pip freeze --all
integration: pip==23.3.2
integration: commands_pre[0]> poetry run pip -V
integration: Exception running subprocess [Errno 2] No such file or directory: 'poetry'
integration: exit 2 (0.00 seconds) /home/fedora/jenkins/workspace/ubipop-pulp-integration> poetry run pip -V
  integration: FAIL code 2 (1.21=setup[1.21]+cmd[0.00] seconds)
  evaluation failed :( (1.27 seconds)
```
I guess the installation of poetry needs to happen in jenkins job. (Similar to gh actions where it's installed before running tox). So far it likely relied on tox installing it.

rbikar:
@jihupdc 
looks like there is a problem with pub--qe instance, poetry install and run work OK.

jihupdc:
> @jihupdc looks like there is a problem with pub--qe instance, poetry install and run work OK.

Thanks for reminder, i'll check ~

rbikar (approved):

